PR #46, opened 2021-04-12
Feat: --debug flag

PopoSensei:
Adds
- debug flag (`--debug`) to enable debug mode for more verbose logging. 
  - Default output is non-coloured LTSV to stderr
- Viper config debugging if debug mode set on with password anonymized to make it less likely to leak
- contexted logger to the executor, allowing debug logging for commands. 
  - Currently implemented for  `router show`. if this looks good to everyone, next MR should add logging for other commands as well.
  
Changes:
- `--color`flag changed to `--force-colours' and `--no-colours`, as we really need the old 'auto' option as well and in this case it's no flags. That is; without flags colors will be on *if* stdout is a terminal but otherwise not.
- spelling of `color` as `colour` across the codebase


internal/config/config.go
```diff
@@ -49,9 +50,11 @@ func New() *Config {
 // GlobalFlags holds information on the flags shared among all commands
 type GlobalFlags struct {
 	ConfigFile    string        `valid:"-"`
-	OutputFormat  string        `valid:"in(human|json|yaml)"`
-	Colors        bool          `valid:"-"`
 	ClientTimeout time.Duration `valid:"-"`
+	Colors        bool          `valid:"-"`
+	Debug         bool          `valid:"-"`
+	OutputFormat  string        `valid:"in(human|json|yaml)"`
+	Wait          bool          `valid:"-"`
```
PopoSensei, line 57: This might've beeb added with merge. Is it wrongly here?

go.sum
```diff
@@ -22,6 +22,8 @@ github.com/adrg/xdg v0.3.2 h1:GUSGQ5pHdev83AYhDSS1A/CX+0JIsxbiWtow2DSA+RU=
 github.com/adrg/xdg v0.3.2/go.mod h1:7I2hH/IT30IsupOpKZ5ue7/qNi3CoKzD6tL3HwpaRMQ=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
+github.com/ansel1/merry v1.5.0 h1:9E6Jj5Yda9SfY3dhbOZUq1CeHl5JA1+mItYvKqokJ7s=
+github.com/ansel1/merry v1.5.0/go.mod h1:wUy/yW0JX0ix9GYvUbciq+bi3jW/vlKPlbpI7qdZpOw=
```
PopoSensei, line 26: Hmmm... Unsure if there is something wrong with the go.sum, since idk how it works in practise. There seems to be too many changes IMO.

internal/config/config.go
```diff
@@ -93,6 +95,10 @@ func (s *Config) Load() error {
 
 	v.Set("config", v.ConfigFileUsed())
 
+	if s.GlobalFlags.Debug {
+		debug.Debug("Load cli", "User config after loading and setting up service")
+		v.Debug()
```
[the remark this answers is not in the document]
PopoSensei: Changed this since the `viper.debug()` is a dump of viper config, not really having anything to do with the `log.debug()` also flume handles debug level, so it should now be better overall. 

current:
```
	log.Debug("Config: Load()", "Viper config logs", v.AllSettings())
```

go.sum
moitias: these are most likely flume dependencies, so probably works as intended. if you want, you can revalidate this by reverting (your local) copy of go.mod and go.sum to what the parent was and then running `go build` (or practically any other relevant command, see [here](https://golang.org/ref/mod#go-mod-file-updates)).

internal/commands/runcommand.go
```diff
@@ -6,17 +6,18 @@ import (
...
 	internal "github.com/UpCloudLtd/upcloud-cli/internal/service"
 	"github.com/spf13/cobra"
 )
 
 func commandRunE(command Command, config *config.Config) func(cmd *cobra.Command, args []string) error {
-
 	switch typedCommand := command.(type) {
 	case NoArgumentCommand:
 		return func(cmd *cobra.Command, args []string) error {
+			log.Debug("runcommand.go", "commandRunE() NoArgumentCommand", "Cmd path:", cmd.CommandPath(), "Arguments:", args)
```
moitias, line 20: not sure if the 'scope' here should be the filename? 

We can get the filename + line automatically with [`runtime.Caller()`](https://golang.org/pkg/runtime/#Caller) so if we want to use it as the scope, would suggest adding some sort of an automatic way in order to reduce possible developer errors (for example, moving some code around).

internal/config/config.go
moitias: Wait at least has been removed at some point, otherwise seems fine?

internal/config/config.go
```diff
@@ -91,6 +94,7 @@ func (s *Config) Load() error {
 
 	v.Set("config", v.ConfigFileUsed())
 
+	log.Debug("Config: Load()", "Viper config logs", v.AllSettings())
```
moitias, line 97: Or; at least let's make the scope consistent?

internal/log/log.go
```diff
@@ -0,0 +1,42 @@
...
+		}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Info Writes info level logs
+func Info(nameSpace string, msg string, args ...interface{}) {
+	log.With("namespace: ", nameSpace).Info(msg, args)
```
moitias, line 36: Is this a flume 'key/value-logging argument'? Eg. does this end up in, say, JSON formatted logs as `{ "namespace: ": "[nameSpace]" }`? I guess the encoding should depend on the output selected as well, or? Not sure, tbh, and that might open up a whole new can of worms.. :o

aakso: AFAIK with flume you should treat the `args ...interface{}` without any formatting. I'd remove spaces and `:` chars and let flume format the kv-pairs. You can of course implement custom encoder for human readable output.

internal/commands/runcommand.go
aakso: This is not very idiomatic flume usage. One should create a flume instance per package or per logical part of the application. That way one can control the logging level per each logger instance.

aakso: Moreover, flume already supports adding caller information with a relevant option so logging the filename is redundant.

PopoSensei: So hmm... How should we init the logger in a way that it should be easy to use for any component? I am starting to think my tryouts here are pretty useless 😅 which is ok, still learning to use go and so on. Since the AddCaller wouldn't work here, since the only path it would show is the log.go pckg

Did think about having the pckg:s create a flume instance with configs. Then the debugMode would be set in the log.go pckg to be called when creating the logger instance. Though unsure if that is good or bad way of doing it and seems a bit silly to even use a different pckg just for debugMode.

So any ideas are welcome :)

PopoSensei: Changed the thought process. And each pckg creates the flume logger and then passes it to the log.go pckg. Though still the caller is in form `@:orig-pckg@log/log.go:39`

internal/log/log.go
PopoSensei: Creating the flume in pckg level. We can input the pckg name in these so the namespace IMO becomes obsolete. We can elaborate more on the location in msg field. I agree with aakso about the formatting of the args. Looks good the way Flume does it.

internal/commands/runcommand.go
```diff
@@ -13,11 +15,15 @@ import (
...
+var (
+	logger = flume.New("runcommand")
+)
+
 func commandRunE(command Command, config *config.Config) func(cmd *cobra.Command, args []string) error {
 	switch typedCommand := command.(type) {
 	case NoArgumentCommand:
 		return func(cmd *cobra.Command, args []string) error {
-			log.Debug("runcommand.go", "commandRunE() NoArgumentCommand", "Cmd path:", cmd.CommandPath(), "Arguments:", args)
+			log.Debug(logger, cmd.CommandPath(), "Arguments:", args)
```
moitias, line 26: Not a big fan of this signature, tbh. Maybe just change our internal logging package to be a wrapper around flume, if we really need this fine-grained control (I'm not convinced we do, as we really only have two scopes - the plumbing and the command being ran)? eg, something like;
```
type Logger struct {
   flume.Logger
}
func New(scope string) *Logger {
  return &Logger{ flume.New(scope) }
}
func (l *Logger) Debug(msg string, args ...interface{}) {
  l.Logger.Debug(msg,args)
}
func (l *Logger) Debug(msg string, args ...interface{}) {
  l.Logger.Debug(msg,args)
}
```
But then again, that raises the question of why not just use flume directly? if flume.Configure applies to all flume instances, I guess we could just implement the level logic somewhere like commmands and go with that?

PopoSensei: Yeah. So we could just in bootstrap.go declare the flume config like this?
``` golang
			flume.Configure(flume.Config{
				AddCaller:    &conf.GlobalFlags.Debug,
				DefaultLevel: flume.DebugLevel,
				Encoding:     "term-color",
			})
			if !conf.GlobalFlags.Debug {
				flume.Configure(flume.Config{
					AddCaller:    &conf.GlobalFlags.Debug,
					DefaultLevel: flume.InfoLevel,
					Encoding:     "term-color",
				})
			}
```

And just in the pckgs init flume and use it. I think you're right and should be enough for now at least.

PopoSensei: This way we also get the caller from flume right `@:runcommand@commands/runcommand.go:47`

internal/commands/runcommand.go
aakso: You can add skipcaller option to skip the wrapping pkg.

moitias:
Finally got around to looking at this again - sorry for the delay - and playing around with it a little bit. 

Looks very good overall, but a few bigger concerns/questions came up; 

- should we redirect the debug log to stderr instead of stdout? Having both the output and log in stdout makes it a little bit difficult with, for example, json output. Not sure if this is an actual problem (i guess --debug would be ran manually and not in a script where the output validity matters anyway?), and/or if we'd actually prefer them in the same fd to make it easy for the users to get all the output in one go? Should we support something like `--debug [filename]` to work around this? I have no real answer to this, curious to hear other views!

- term color being always on is a bit of a problem, as that will generate ansi codes in piping output, that make the logs pretty difficult to read outside of a terminal. I'm not sure how easily we can control this* and whether there is an actual use for color in the logs. It's nice to get colorized logs, but our use case for the logs (eg. debug use only) is a bit different than the common 'server logs' case where, I feel, colors make more sense?

*) we do have a terminal package which seems to have relevant methods to query it but I'm not sure if those work or not, as piping the humanized output also results in ansi codes in the logs..

internal/config/config.go
```diff
@@ -91,6 +95,7 @@ func (s *Config) Load() error {
 
 	v.Set("config", v.ConfigFileUsed())
 
+	logger.Debug("Viper configs", v.AllSettings())
```
moitias, line 98: do note that this will log, among other things, user password:
```
matias@kiri ~/c/upctl (feat/debug-flag)> 
go run cmd/upctl/main.go server show 00cd5b9f-4996-4374-86f3-bbcb7f650c68 --debug > foo && head foo
2021-04-29T11:15:31.894+0300 DBG | Viper configs  	value:map[client-timeout:1m0s colours:true config:/home/matias/.config/upctl.yaml debug:true output:human password:REMOVED username:devmatiasapi]	@:config@config/config.go:98
```

internal/commands/executor.go
```diff
@@ -24,6 +27,10 @@ type Executor interface {
 	Account() service.Account
 	Plan() service.Plans
 	All() internal.AllServices
+	Debug(msg string, args ...interface{})
+	Info(msg string, args ...interface{})
+	Error(msg string, args ...interface{})
+	WithLogger(args ...interface{}) Executor
```
moitias, line 33: Not sure about these, but went with 'as close to flume as possible'. Error(), for example, doesn't make it obviously clear that it's a error *logging* method. Maybe DebugLog / LogDebug etc? Or maybe executor could just provide Logger() flume.logger method?

internal/core/core.go
```diff
@@ -25,7 +28,29 @@ func BuildRootCmd(conf *config.Config) cobra.Command {
...
+			terminal.ForceColours(conf.GlobalFlags.Colors && terminal.IsStdoutTerminal())
+
+			// Set up flume
+			flume.SetOut(os.Stderr)
+			// TODO: should we make the level configurable?
+			logLvl := flume.DebugLevel
+			if !conf.GlobalFlags.Debug {
+				// not debugging, no log output!
+				logLvl = flume.OffLevel
+			}
```
moitias, line 40: Not really if debug log levels even make sense in this context? Could we just make do with one level of logging - it would make the executor side simpler (only Debug/DebugLog method)?

internal/commands/executor.go
```diff
@@ -96,10 +121,11 @@ func (e executorImpl) All() internal.AllServices {
 }
 
 // NewExecutor creates the default Executor
-func NewExecutor(cfg *config.Config, svc internal.AllServices) Executor {
+func NewExecutor(cfg *config.Config, svc internal.AllServices, logger flume.Logger) Executor {
```
moitias, line 124: Added the logger to the executor, rather than changing the execute() signatures all over. Not sure if this is the best option?

internal/core/core.go
aakso: agreed, it makes no sense to have different levels if `--debug` enables the logging in the first place.

internal/core/core.go
```diff
@@ -25,7 +28,29 @@ func BuildRootCmd(conf *config.Config) cobra.Command {
 				return err
 			}
 
-			terminal.ForceColours(conf.GlobalFlags.Colors)
+			terminal.ForceColours(conf.GlobalFlags.Colors && terminal.IsStdoutTerminal())
```
aakso, line 31: Colours: there is a wider problem here with them. Somewhere along the way it seems that GlobalFlags.Colors (wrong spelling btw as company policy is British English) became authoritative setting of determining whether to display colours. The call to `terminal.Colours() bool` already gives an educated guess whether colours should be enabled (by examining if the stdout is a tty). 

Discussed this with @moitias and one possibility is that GlobalFlags.Colous would be removed and we'd introduce two new booleans `ForceColours` and `DisableColours` with relevant flags. We would then only call terminal.ForceColours should one of the flags be set and otherwise let automatic detection of terminal enable the colours.

When it comes to logging I would suggest going without any colours at all especially if we log to stderr.

aakso (commented):
This is far better approach 👍

internal/core/core.go
moitias: :+1: these have been now implemented in the head of the branch.

PR #47, opened 2021-04-13
Fix module path to match repository name

Darep:
Fixes:

```
go get: github.com/UpCloudLtd/upcloud-cli@v0.1.1: parsing go.mod:
	module declares its path as: github.com/UpCloudLtd/cli
	        but was required as: github.com/UpCloudLtd/upcloud-cli
```

### DIFF
--- a/cmd/upctl/main.go
+++ b/cmd/upctl/main.go
@@ -3,9 +3,9 @@ package main
 import (
 	"os"
 
-	"github.com/UpCloudLtd/cli/internal/commands/all"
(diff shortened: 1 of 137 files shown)


moitias:
Looks like the change is still partial, as the tests fail? Otherwise, definitely :+1: for this!

Darep:
> Looks like the change is still partial, as the tests fail? Otherwise, definitely 👍 for this!

Fixed now! (dunno why that one file got left behind 😅 )

PR #49, opened 2021-04-13
Doc: add a CI to auto generate commands documentation

Darep:
This MR:

- adds a script for auto-generating Markdown documentation
- Add a CI that pushes the generated doc in master whenever possible


moitias:
Looks gooood! I guess the only question if we really want to commit the generated files to the repository? eg. do we want to host the docs on github as well or figure out a better way to do it? Not really against commiting, but then we should make sure that gets triggered on the CI build chain as well (and commits those changes as well?? :o)

Darep:
> Looks gooood! I guess the only question if we really want to commit the generated files to the repository? eg. do we want to host the docs on github as well or figure out a better way to do it? Not really against commiting, but then we should make sure that gets triggered on the CI build chain as well (and commits those changes as well?? :o)

Yep, agree! Need to figure out how we want to do this 😄  One solution: add `gh-pages` branch, put the script & generated docs in there,  link to that from README. That branch can then run separately automatically & pollute that branch with commits, so it doesn't interfere with our master 🤔

kaminek:
I think that this should remain in the git repos, of course the CI will manage this. but yeah if we have a good idea how makeup this markdown in a better way, it could be a good plus

generate_docs.go
```diff
@@ -0,0 +1,23 @@
+package main
```
kaminek, line 1: this pkg should be in cmd/doc-gen/main.go and treated as a standalone command

docs/upctl.md
```diff
@@ -0,0 +1,35 @@
+## upctl
```
kaminek, line 1: We should have a reference to this file from the main README.md

kaminek:
there is a missing dependency cf golanci-lint

kaminek:
a CI with a bot (UpCloud bot user) should be added to regenerate the doc whenever a MR is open and if doc is being updated the bot adds a commit with the doc diff to the MR.

Darep:
> a CI with a bot (UpCloud bot user) should be added to regenerate the doc whenever a MR is open and if doc is being updated the bot adds a commit with the doc diff to the MR.

Yeah -- GitHub actions maybe? 🤔

README.md
```diff
@@ -6,20 +6,34 @@
...
 server      Manage servers
 storage     Manage storages
+version     Display software infomation
+
+Options:
+  -t, --client-timeout duration   CLI timeout when using interactive mode on some commands
+                                  Default: 1m0s
+
+  --colours bool                  Use terminal colours (supported: auto, true, false)
+                                  Default: true
```
moitias, line 31: default: auto, i guess?

README.md
```diff
@@ -151,14 +170,15 @@ Use `make` to build and test the CLI. Makefile help can be found:
...
-build-darwin      Build program binary for darwin x86_64
-build-windows     Build program binary for windows x86_64
-test              Run tests
-fmt               Run gofmt on all source files
-clean             Cleanup everything
+build                Build program binary for current os/arch
+doc                  Build documentation (markdown)
+build-all            Build all targets
+build-linux          Build program binary for linux x86_64
+build-darwin         Build program binary for darwin x86_64
```
moitias, line 177: we're supporting darwin/x86_64 and darwin/arm64 now, correct? If so, these should probably reflect it. 

(Although; that does raise another question - what about, say, linux/arm64?)

docs/upctl.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Options
+
+```
+  -t, --client-timeout duration   CLI timeout when using interactive mode on some commands (default 1m0s)
+      --colours                   Use terminal colours (supported: auto, true, false) (default true)
+      --config string             Config file
+  -h, --help                      help for upctl
+  -o, --output string             Output format (supported: json, yaml and human) (default "human")
+  -w, --wait                      Wait for the command to be completed
```
moitias, line 21: looks like wait snuck back in from somewhere?

docs/upctl_ip-address_assign.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Options
+
+```
+      --access string   Is address for utility or public network. (default "public")
+      --family string   The address family of new IP address. (default "IPv4")
+      --server string   The server the ip address is assigned to.
+      --mac string      MAC address of server interface to assign address to. Required for non-floating addresses.
+      --zone string     Zone of address, required when assigning a detached floating IP address.
+      --floating        Whether the address to be assigned is a floating one.
```
moitias, line 17: this is actually a bit confusing now, i think? 

Broader question: should we, before 1.0, have another go at the 'tri-state' boolean flags, (eg. true, false or not set) and turning them into --enable-floating / --disable-floating? I think this was discussed earlier on and there was even a bit of poc code floating around in the discussion.

docs/upctl_ip-address_modify.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+upctl ip-address modify <Address/PTRRecord...> [flags]
+```
+
+### Options
+
+```
+      --mac string          MAC address of server interface to attach floating IP to.
+      --ptr-record string   A fully qualified domain name.
```
moitias, line 13: probably should make it explicit that this flag is the *new* ptr record for the ip-address? it's not immediately clear that this is the update and not just a way of referring to it.

docs/upctl_network_create.md
```diff
@@ -0,0 +1,45 @@
...
+                                 
+                                  Fields: 
+                                   address: string 
+                                   family: string 
+                                   gateway: string 
+                                   dhcp: true/false 
+                                   dhcp-default-route: true/false 
+                                   dhcp-dns: array of strings 
+                                 Usage: 
+                                 	--ip-network 'address=94.23.112.143,"dhcp-dns=<value1>,<value2>",gateway=<gateway>,dhcp=true'
```
moitias, line 25: Are these generated from the examples done earlier? Would be nice to get these wrapped in the 'console output quotes' (backquotes?) in the markdown output if it's possinle.

docs/upctl_router_list.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+upctl router list [flags]
+```
+
+### Options
+
+```
+      --all       Show all routers.
+      --normal    Show normal routers (default). (default true)
```
moitias, line 13: stuttering a bit here

moitias: also; --normal is kind of a weird flag all in all and .. i'm not sure if this flag has any use? when would someone use it and how?

docs/upctl_server_create.md
```diff
@@ -0,0 +1,57 @@
...
+### Options
+
+```
+      --avoid-host int                  Use this to make sure VMs do not reside on specific host. Refers to value from host -attribute. Useful when building HA-environments.
+      --host int                        Use this to start a VM on a specific host. Refers to value from host -attribute. Only available for private cloud hosts.
+      --boot-order string               The boot device order, disk / cdrom / network or comma separated combination.
+      --user-data string                Defines URL for a server setup script, or the script body itself.
+      --cores int                       Number of cores. Use only when defining a flexible ("custom") plan.
+      --memory int                      Memory amount in MiB. Use only when defining a flexible ("custom") plan.
+      --title string                    Visible name.
```
moitias, line 18: visible where?

docs/upctl_server_create.md
```diff
@@ -0,0 +1,57 @@
...
+
+```
+upctl server create [flags]
+```
+
+### Options
+
+```
+      --avoid-host int                  Use this to make sure VMs do not reside on specific host. Refers to value from host -attribute. Useful when building HA-environments.
+      --host int                        Use this to start a VM on a specific host. Refers to value from host -attribute. Only available for private cloud hosts.
```
moitias, line 13: what is the host attribute? this is unclear

docs/upctl_server_create.md
```diff
@@ -0,0 +1,57 @@
...
+      --memory int                      Memory amount in MiB. Use only when defining a flexible ("custom") plan.
+      --title string                    Visible name.
+      --hostname string                 Hostname.
+      --plan string                     Server plan name. See "server plans" command for valid plans. Set --cores and --memory for a flexible plan. (default "1xCPU-2GB")
+      --os string                       Server OS to use (will be the first storage device). Set to empty to fully customise the storages. (default "Ubuntu Server 20.04 LTS (Focal Fossa)")
+      --os-storage-size os              OS storage size in GiB. This is only applicable if os is also set. Zero value makes the disk equal to the minimum size of the template.
+      --zone string                     Zone where to create the server.
+      --password-delivery string        Defines how password is delivered. Available: email, sms (default "none")
+      --simple-backup string            Simple backup rule. Format (HHMM,{dailies,weeklies,monthlies}). Example: 2300,dailies
+      --time-zone string                Time zone to set the RTC to. (default "UTC")
```
moitias, line 26: RTC could be made more clear, I think? (Maybe something like RTC/VM system time?)

docs/upctl_server_modify.md
```diff
@@ -0,0 +1,47 @@
...
+Modifies the configuration of an existing server
+
+```
+upctl server modify <UUID/Title/Hostname...> [flags]
+```
+
+### Options
+
+```
+      --boot-order string               The boot device order.
```
moitias, line 12: This could use an example of two?

docs/upctl_server_network-interface_create.md
```diff
@@ -0,0 +1,38 @@
...
+### Options
+
+```
+      --network string         Virtual network ID or name to join.
+      --type string            Set the type of the network. Available: public, utility, private (default "private")
+      --family string          The address family of new IP address. (default "IPv4")
+      --index int              Interface index.
+      --bootable               Whether to try booting through the interface.
+      --source-ip-filtering    Whether source IP filtering is enabled on the interface. Disabling it is allowed only for SDN private interfaces.
+      --ip-addresses strings   Array of IP addresses, multiple can be declared
```
moitias, line 18: not an array. it's a comma-separated list, at least judging by the example below

docs/upctl_server_network-interface_modify.md
```diff
@@ -0,0 +1,35 @@
...
+
+```
+upctl server network-interface modify <UUID/Title/Hostname...> [flags]
+```
+
+### Options
+
+```
+      --index int                    Index of the interface to modify.
+      --new-index int                Index of the interface to modify.
```
moitias, line 13: Euhm. 'new index to move the interface to?' or something?

docs/upctl_server_network-interface_modify.md
```diff
@@ -0,0 +1,35 @@
...
+      --bootable string              Whether to try booting through the interface.
+      --source-ip-filtering string   Whether source IP filtering is enabled on the interface. Disabling it is allowed only for SDN private interfaces.
```
moitias, line 15: looks like these two should be bools (tristate?) but are actually strings.. i guess the bool flags could really use some love before release?

docs/upctl_ip-address_assign.md
kaminek: I don't think there are tree states as the struct get initalized with a False value, https://github.com/UpCloudLtd/upcloud-go-api/blob/master/upcloud/ip_address.go#L52

docs/upctl_server_restart.md
```diff
@@ -0,0 +1,30 @@
...
+Restart a server
+
+```
+upctl server restart <UUID/Title/Hostname...> [flags]
+```
+
+### Options
+
+```
+      --stop-type string   Restart type. Available: soft, hard (default "soft")
```
moitias, line 12: should this be --restart-type? i guess stop isnt too bad either but the documentation could then refer to 'stop type', maybe?

docs/upctl_server_storage_attach.md
```diff
@@ -0,0 +1,34 @@
...
+```
+upctl server storage attach <UUID/Title/Hostname...> [flags]
+```
+
+### Options
+
+```
+      --type string      Type of the attached storage. Available: disk, cdrom (default "disk")
+      --address string   Address where the storage device is attached on the server. 
+                         Specify only the bus name (ide/scsi/virtio) to auto-select next available address from that bus. (default "virtio")
```
moitias, line 14: could also specify how the normal addresses are input; `virtio:1` afaik?

docs/upctl_server_storage_detach.md
```diff
@@ -0,0 +1,30 @@
...
+Detaches a storage resource from a server
+
+```
+upctl server storage detach <UUID/Title/Hostname...> [flags]
+```
+
+### Options
+
+```
+      --address string   Detach the storage attached to this address.
```
moitias, line 12: this would be required, then? should we document the fact that this (and any other similar) flag is required?

docs/upctl_storage_create.md
```diff
@@ -0,0 +1,37 @@
...
+upctl storage create [flags]
+```
+
+### Options
+
+```
+      --title string             Storage title.
+      --size int                 Size of the storage in GiB. (default 10)
+      --zone string              Physical location of the storage. See zone listing for valid zones.
+      --tier string              Storage tier. (default "maxiops")
```
moitias, line 15: other options should probably be listed here as well

docs/upctl.md
Darep: The branch was just behind & the docs were generated from older code, now up-to-date :)

README.md
kaminek: yeah moving to another PR

PR #50, opened 2021-04-13
Refactor creation of command runE to unify execution of all types of commands

moitias:
The previous refactoring related to argument handling introduced a regression; other execution modes except the multiple/parallel mode would not create any output. 

This MR refactors how the RunE method is generated for the different types of argument handling commands. Plenty of bits were DRYed up and the separate bits have been split apart to logical pieces.


kaminek:
didn't found the issue with go-critic, maybe open a github Issue

PR #55, opened 2021-04-14
fix(root): use default cobra behavious when called

kaminek:


moitias:
could use a little description of *why* this was done, but I'm guessing it relates to showing usage when running just 'upctl'? If so, LGTM :+1:

PR #57, opened 2021-04-16
Refactor runcommand, again

moitias:
This PR refactors runcommand to accept service from outside, in order to facilitate testing the runcommand logic. Previously, service was created within execute() which made it impossible to pass in mock service. 

Also makes the runcommand a bit more straightforward with less return func()s and adds the forgotten golangci config that disables just the unlambda test (which still panics with this code).


kaminek:
LGTM

PR #61, opened 2021-04-19
Refactor runcommand some more

moitias:
Separate more bits out of the runcommand internals; move render to be called by runcommandE and move executor creation to runcommandE, as well. This is in order to separate the concerns better and make the separate bits more testable.


kaminek:
LGTM

PR #62, opened 2021-04-19
Add more command examples

ka-myl:


docs/upctl_server_create.md
```diff
@@ -6,6 +6,14 @@ Create a server
 upctl server create [flags]
 ```
 
+### Examples
+
+```
+upctl server create --title my_server --zone fi-hel1 --hostname myapp --password-delivery email
```
Darep, line 12: hmm, does the CLI accept e.g. `--title "My Server"`, could hilight that in the example 🤔

internal/commands/server/stop.go
```diff
@@ -15,8 +15,14 @@ import (
 
 // StopCommand creates the "server stop" command
 func StopCommand() commands.Command {
+	examples := fmt.Sprint(
+		"upctl server stop my_server\n",
+		"upctl server stop my_server1 my_server2",
+		"upctl server stop 00cbe2f3-4cf9-408b-afee-bd340e13cdd8",
```
Darep, line 21: I would promote always using the UUID as it's less error prone :smile: Like, put two examples using UUID first, and one last example with title 😄

Darep (approved):
Much better than before, so feel free to merge! LGTM 👍 

P.S. I did not test all the examples to be working 😉

internal/commands/server/stop.go
ka-myl: The thing about uuids is that it's not really possible to type them by hand, you would have to check them in a control panel. Which is ok when you have some automation scripts around upctl, but if you are just typing stuff by hand - not so much :P

But overall yeah, let's have some consistent approach here - UUID examples first, then name/title/whatever

PR #64, opened 2021-04-19
fix(docs): fix mac and zone flag descriptions for ip-address assign

ka-myl:
Not 100% sure, but base on usage I believe this is actually the correct description:
image


moitias:
LGTM :+1:

kaminek:
Hum missing server uuid in the livelog no?

ka-myl:
@kaminek Added a separate ticket for it :)

PR #65, opened 2021-04-19
Chore/more command examples

ka-myl:
Did not regenerate files because, as far as I understand, CI should handle it from now on

### DIFF
--- a/internal/commands/network/create.go
+++ b/internal/commands/network/create.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+
 	"github.com/UpCloudLtd/upcloud-cli/internal/commands"
 	"github.com/UpCloudLtd/upcloud-cli/internal/commands/ipaddress"
 	"github.com/UpCloudLtd/upcloud-cli/internal/output"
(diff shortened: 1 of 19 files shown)


internal/commands/network/create.go
```diff
@@ -25,7 +26,12 @@ var _ commands.Command = &createCommand{}
 // CreateCommand creates the 'network create' command
 func CreateCommand() commands.Command {
 	return &createCommand{
-		BaseCommand: commands.New("create", "Create a network", ""),
+		BaseCommand: commands.New(
+			"create",
+			"Create a network",
+			"upctl network create --name \"My Network\" --zone pl-waw1 --ip-network address=10.0.1.0/24",
```
moitias, line 32: Very minor, but maybe we could use backquotes here to not have to escape the quotes?

internal/commands/network/create.go
```diff
@@ -43,11 +49,8 @@ func (s *createCommand) InitCommand() {
 		"  gateway: string \n"+
 		"  dhcp: true/false \n"+
 		"  dhcp-default-route: true/false \n"+
-		"  dhcp-dns: array of strings \n"+
-		"Usage: \n"+
-		"	--ip-network 'address=94.23.112.143,\"dhcp-dns=<value1>,<value2>\",gateway=<gateway>,dhcp=true' \n"+
-		"	--ip-network address=94.43.112.143/32,dhcp-dns=<value>\n")
-	s.AddFlags(fs) // TODO(ana): replace usage with examples once the refactor is done.
+		"  dhcp-dns: array of strings \n")
```
moitias, line 52: is the tailing newline necessary/standard? This could also be backquoted, I guess, although that would lead to kind of an ugly look..

moitias:
LGTM :+1:

internal/commands/network/create.go
ka-myl: Honestly not sure, we could probably delete those new lines as they don't work anyway (everything prints in one line in my terminal)... But that sounds like a separate MR, since there's a number of those that would need fixing

internal/commands/network/create.go
ka-myl: Done

PR #66, opened 2021-04-20
Chore/add missing commands examples

ka-myl:
Looks like the commands added in [this PR](https://github.com/UpCloudLtd/upcloud-cli/pull/65) somehow did not end up in master.
I just copied them manually since that was fastest :P

### DIFF
--- a/internal/commands/network/modify.go
+++ b/internal/commands/network/modify.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+
 	"github.com/UpCloudLtd/upcloud-cli/internal/commands"
 	"github.com/UpCloudLtd/upcloud-cli/internal/completion"
 	"github.com/UpCloudLtd/upcloud-cli/internal/output"
@@ -24,7 +25,12 @@ type modifyCommand struct {
 // ModifyCommand creates the "network modify" command
 func ModifyCommand() commands.Command {
 	return &modifyCommand{
-		BaseCommand: commands.New("modify", "Modify a network", ""),
+		BaseCommand: commands.New(
+			"modify",
+			"Modify a network",
+			"upctl network modify 037a530b-533e-4cef-b6ad-6af8094bb2bc --ip-network dhcp=false,family=IPv4",
+			`upctl network modify "My Network" --name "My Super Network"`,
+		),
 	}
 }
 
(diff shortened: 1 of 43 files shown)


internal/commands/network/modify.go
```diff
@@ -39,11 +45,8 @@ func (s *modifyCommand) InitCommand() {
 		"  gateway: string \n"+
 		"  dhcp: true/false \n"+
 		"  dhcp-default-route: true/false \n"+
-		"  dhcp-dns: array of strings \n"+
-		"Usage \n"+
-		"	--ip-network dhcp-dns=<value1>,family=IPv4 \n"+
-		" --ip-network 'dhcp=true,\"dhcp-dns=<value1>,<value2>\",family=IPv6'")
-	s.AddFlags(fs) // TODO(ana): replace usage with examples once the refactor is done.
+		"  dhcp-dns: array of strings \n")
```
moitias, line 48: is the tailing newline necessary? eg. does it create an empty line between the flag descriptions and if so, is that what we want?

internal/commands/serverstorage/attach.go
```diff
@@ -29,7 +29,12 @@ type attachParams struct {
 // AttachCommand creates the "server storage attach" command
 func AttachCommand() commands.Command {
 	return &attachCommand{
-		BaseCommand: commands.New("attach", "Attach a storage as a device to a server", ""),
+		BaseCommand: commands.New(
+			"attach",
+			"Attach a storage as a device to a server",
+			"upctl server storage attach 00038afc-d526-4148-af0e-d2f1eeaded9b --storage 015899e0-0a68-4949-85bb-261a99de5fdd",
+			`upctl server storage attach my_server1 --storage "My Storage"`,
```
moitias, line 36: maybe could add examples of --address use here as well? I had problems understanding it to begin with. The flag description has since been clarified but might make sense to add examples as well?

moitias:
LGTM :+1:

internal/commands/network/modify.go
ka-myl: As far as I can tell I doesn't do anything (at least on mac, zsh). All those newlines in this description seem to not have any effect

internal/commands/serverstorage/attach.go
ka-myl: done :)

kaminek:
LGTM

internal/commands/network/modify.go
moitias: Interesting, good catch. Seems like something has changed somewhere at some point.. Ticketing this.

PR #71, opened 2021-04-21
fix(storage): size up os disk to the correct value

kaminek:
We need to set the correct disk size:
* Classic Plan: follow the plan disk size if not overrided with custom
  --os-storage-size
* Custom Plan: use OS default size if not specified over the min size allowed
  (10G)


moitias:
LGTM :+1:

PR #72, opened 2021-04-22
fix(client-timeout): increase to raisonnable value

kaminek:


kaminek:
won't d, duplicated

PR #74, opened 2021-04-27
Add colors to some missing UUIDs

kaminek:
and fix a empty display

### DIFF
--- a/internal/commands/ipaddress/show.go
+++ b/internal/commands/ipaddress/show.go
@@ -47,7 +47,11 @@ func (s *showCommand) Execute(exec commands.Executor, arg string) (output.Output
 					{Title: "Family:", Key: "access", Value: ipAddress.Family},
 					{Title: "Part of Plan:", Key: "access", Value: ipAddress.PartOfPlan, Format: output.BoolFormat},
 					{Title: "PTR Record:", Key: "access", Value: ipAddress.PTRRecord},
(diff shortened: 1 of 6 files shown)


internal/commands/ipaddress/show.go
```diff
@@ -47,7 +47,11 @@ func (s *showCommand) Execute(exec commands.Executor, arg string) (output.Output
 					{Title: "Family:", Key: "access", Value: ipAddress.Family},
 					{Title: "Part of Plan:", Key: "access", Value: ipAddress.PartOfPlan, Format: output.BoolFormat},
 					{Title: "PTR Record:", Key: "access", Value: ipAddress.PTRRecord},
-					{Title: "Server UUID:", Key: "access", Value: ipAddress.ServerUUID},
+					{
+						Title: "Server UUID:",
+						Key:   "access", Value: ipAddress.ServerUUID,
```
moitias, line 52: The keys get used in json/yaml output so probably dont want to name them all the same... 

(sidenote: we should go through all of these i guess?)

moitias: (or; should we validate this even when rendering?)

kaminek: :scream: didn't even noticed it, yes it should be renamed properly

moitias:
Besides the keys (which, I guess, could be another MR?), LGTM :+1:

PR #75, opened 2021-04-28
feat(flags/boolean): implement tristate boolean flags

moitias:


internal/commands/server/modify.go
```diff
@@ -2,14 +2,13 @@ package server
 
 import (
 	"fmt"
-
```
kaminek, line 5: please keep the separation

kaminek (commented):
LGTM :+1:

internal/commands/server/modify.go
moitias: oops, mb! will do. 

Although - we really should figure out a way to let the linter do this for us..

PR #78, opened 2021-04-29
feat: implement initial terminal width support

moitias:
- add terminal width detection 
- set up livelog rendering to respect terminal width and handle resize events (not on Windows)

Initially, the aim was to add terminal width support to table rendering as well, but it's not clear how we should handle that case as when the table does not fit, truncating the table does not seem like a good idea and when it's too small, the library used does not really give us easy ways of fitting to available width.

Thus, we just render tables like we used to - at the minimum width that we can fit the table into without truncating values.


kaminek (approved):
LGTM

PR #79, opened 2021-04-29
Dockerised version

ajmyyra:
This PR brings in a smallish Dockerfile build and creates a build step for a suitable binary. Once merged, we can have Dockerhub build a new tagged version whenever we publish one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.16-buster as builder
+RUN apt-get -y install ca-certificates
+WORKDIR /go/upctl/
+COPY . .
+RUN make build-dockerised
+
+FROM scratch
+LABEL org.label-schema.vcs-url="https://github.com/UpCloudLtd/upcloud-cli"
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=builder /go/upctl/bin/*dockerised-linux-amd64 /upctl
+
+ENTRYPOINT ["/upctl"]
--- a/Makefile
+++ b/Makefile
@@ -12,12 +12,13 @@ TESTPKGS = $(shell env GO111MODULE=on $(GO) list -f \
 			'{{ if or .TestGoFiles .XTestGoFiles }}{{ .ImportPath }}{{ end }}' \
 			$(PKGS))
 
-BIN_DIR         = $(CURDIR)/bin
-CLI_BIN         = $(CLI)
-DOC_GEN_BIN     = $(DOC_GEN)
-BIN_LINUX       = $(CLI_BIN)-$(VERSION)-linux-amd64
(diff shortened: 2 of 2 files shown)


kaminek:
I think that you need to integrate the docker image in the CD too. checkout https://goreleaser.com/customization/docker/

Makefile
```diff
@@ -45,6 +46,13 @@ build-linux: ; $(info building executable for Linux x86_64…) @ ## Build progra
 		-ldflags '-X $(MODULE)/internal/config.Version=$(VERSION) -X $(MODULE)/internal/config.BuildDate=$(DATE)' \
 		-o $(BIN_DIR)/$(BIN_LINUX) cmd/$(CLI)/main.go
 
+.PHONY: build-dockerised
```
kaminek, line 49: why a new binary, what is different from the one above?

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.16-buster as builder
+RUN mkdir /go/upctl
```
Darep, line 2: `WORKDIR` should create the directory too

Makefile
Darep: `CGO_ENABLED=0` seems to be the only difference, totally new flag for me 🤔 Seems to control the CGo standard libraries (build them vs use prebuilt) or something(?)

Darep:
Gave it a spin locally and seems to work! Image is 10.4MB 😎 Nice 👍 

fyi, tested it like this:

```
docker run --rm -v $(PWD)/upctl.yaml:/.config/upctl.yaml upctl server list
```

Makefile
ajmyyra: CGO_ENABLED=0 and especially -w for LD_FLAGS. Unfortunately the binary won't run under scratch without them.

ajmyyra:
My idea was to use Docker Hub [builds](https://docs.docker.com/docker-hub/builds/) for this one. Any objections on this? If yes, I can of course add the CD step to this end as well.

Dockerfile
ajmyyra: Cheers! Should have remembered but nope. :) Fixed now.

Makefile
ajmyyra: Now with a longer answer as I have the time. CGO_ENABLED=0 creates a statically linked binary, that is needed as the `scratch` container doesn't have any existing libraries. It leads to a bigger binary (10MB vs ~3MB), but it's still a lot smaller than having those in an Alpine container. -w in ldflags turns off debugging information, leading to a smaller binary than what we'd otherwise get.

Darep:
> My idea was to use Docker Hub [builds](https://docs.docker.com/docker-hub/builds/) for this one. Any objections on this? If yes, I can of course add the CD step to this end as well.

Sounds good to me, though i have zero experience with goreleaser and on its benefits :P

PR #81, opened 2021-05-03
fix(config/load): improve error message when loading invalid YAML

### LINKED ISSUE #77: yaml config parse error handling failure
**Environment (required):**
- macOS (M1)
- Version 1.0.0

** bad yaml file parsing - unhandled error**
Error: cannot load configuration: While parsing config: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `usernam...` into map[string]interface {}

**To Reproduce**
1. Create a config file with username:<username> and password:<password> i.e. without the space.
2. Marshalling error is displayed. Not all users will be aware of yaml specs to know that space is to be added

**Expected behavior**
Please let a user know that a space character is expected or provide an example.

**Observed behavior**
Error message prevents user from running upctl including checking version.

moitias:
Fixes #77

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -85,7 +85,7 @@ func (s *Config) Load() error {
 	// Attempt to read the config file, ignoring only config file not found errors
 	if err := v.ReadInConfig(); err != nil {
 		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-			return err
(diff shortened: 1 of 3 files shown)


ajmyyra:
Seems nice, a better & more informative error.

Darep (approved):
LGTM!

PR #86, opened 2021-05-17
chore(lint): enhance linter list

kaminek:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -14,7 +14,6 @@ linters:
     - gocritic
     - gofmt
     - goimports
-    - golint
     - goprintffuncname
 #    - gosec
     - gosimple
@@ -23,6 +22,7 @@ linters:
     - misspell
     - nakedret
     - nolintlint
+    - revive
     - rowserrcheck
     - staticcheck
     - structcheck
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,8 +1,42 @@
+# Golang CI pipeline configuration
+linters:
+  disable-all: true
+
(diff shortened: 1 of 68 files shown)


internal/config/optionalboolean.go
```diff
@@ -20,6 +21,11 @@ type OptionalBoolean int
 // an error when trying to set a non-Unset OptionalBoolean
 
 const (
+	// FalseFlagValue false Bool String value
```
moitias, line 24: The comment could be more descriptive about what it actually does; maybe something like 'FalseFlagValue is the command-line flag string for false values'?

.golangci.yaml
```diff
@@ -1,8 +1,42 @@
...
   gocritic:
     disabled-checks:
       # TODO: re-enable after figuring out why runcommand fails on go-critic..
       - unlambda
-linters:
-  enable:
-    - gocritic
\ No newline at end of file
+  goconst:
+    min-len: 5
```
moitias, line 42: I'm a bit wary of custom settings for the linters* - what's the rationale for changing it? Do we have a lot of strings that would otherwise warn and if so, why not constify them if we want goconst enabled?

*) Two reasons - copypasting custom settings across different codebases will make it easy to leave them unmaintained and consistency across our and all other codebases (although, with goconst, I guess it's not that important)

moitias (commented):
Few comments and a weirdo failing test to resolve but overall LGTM :+1:

.golangci.yaml
kaminek: though that we can keep the golangci-lnit config file consistent across our projects. I just copy paste it from the terraform provider.

PR #89, opened 2021-05-27
fix(network/router): implement router attaching and detaching with new go-api

moitias:
- adds a `--detach-router` toggle flag to `network modify`, that triggers a detach
- changes `--router` to only work for attaching, eg. one can't detach with `--router ""` anymore. (nb. due to the go-api issue, it didnt work before)

Considered adding separate commands for detaching and attaching, but did not like the approach as `network detachrouter` is a bit clunky? Other suggestions for improving/changing UI are welcome.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/UpCloudLtd/upcloud-cli
 go 1.13
 
 require (
-	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20201130170350-7fc02444127a
+	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210525102709-661419a1956b
 	github.com/adrg/xdg v0.3.2
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/fsnotify/fsnotify v1.4.9 // indirect
--- a/internal/commands/network/modify.go
+++ b/internal/commands/network/modify.go
@@ -43,8 +43,8 @@ func ModifyCommand() commands.Command {
 func (s *modifyCommand) InitCommand() {
 	fs := &pflag.FlagSet{}
 	fs.StringVar(&s.name, "name", "", "Set name of the private network.")
-	fs.StringVar(&s.attachRouter, "router", "", "Set the router attachment.")
-	config.AddToggleFlag(fs, &s.detachRouter, "detach-router", false, "Detach the router attachment.")
+	fs.StringVar(&s.attachRouter, "router", "", "Attach a router to this network, specified by router UUID or router name.")
+	config.AddToggleFlag(fs, &s.detachRouter, "detach-router", false, "Detach a router from this network.")
 	fs.StringArrayVar(&s.networks, "ip-network", s.networks, "The ip network with modified values. \n\n"+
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/UpCloudLtd/upcloud-cli
 go 1.13
 
 require (
-	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20201130170350-7fc02444127a
+	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210525102709-661419a1956b
```
kaminek, line 6: same issue for go sdk pkg version

internal/commands/network/modify.go
```diff
@@ -68,21 +77,75 @@ func (s *modifyCommand) ExecuteSingleArgument(exec commands.Executor, arg string
...
-		logline.SetDetails(err.Error(), "error: ")
-		return nil, err
+	var network *upcloud.Network
+	if s.name != "" || len(networks) > 0 {
+		// we want to update name and/or networks
+		logline.SetMessage(fmt.Sprintf("%s: sending modify request", msg))
+		res, err := exec.Network().ModifyNetwork(&request.ModifyNetworkRequest{
+			UUID:       arg,
+			Name:       s.name,
+			Zone:       "", // TODO: should this be implemented?
```
kaminek, line 88: this fields isn't even in the API spec

internal/commands/network/modify.go
```diff
@@ -38,7 +43,8 @@ func ModifyCommand() commands.Command {
 func (s *modifyCommand) InitCommand() {
 	fs := &pflag.FlagSet{}
 	fs.StringVar(&s.name, "name", "", "Set name of the private network.")
-	fs.StringVar(&s.router, "router", "", "Change or clear the router attachment.")
+	fs.StringVar(&s.attachRouter, "router", "", "Set the router attachment.")
```
kaminek, line 46: Set the router attachment UUID/name

kaminek (approved):
a part from the comments, LGTM

internal/commands/network/modify.go
moitias: I wonder if it's a leftover from an admin API? Anyhow, this TODO is a leftover from the very early refactoring days, so I'd still keep it there.

go.mod
moitias: Yeah, this is unfortunately not fixable unless we also change the import paths. We *could* just tag it as `v4.0.0` (instead of `4.0.0`) but that would result in an `incompatible` version tag and I'm not sure that's an actual improvement. see [here](https://golang.org/ref/mod#incompatible-versions)

internal/commands/network/modify.go
moitias: Changed the help text(s) to;
```
fs.StringVar(&s.attachRouter, "router", "", "Attach a router to this network, specified by router UUID or router name.")
config.AddToggleFlag(fs, &s.detachRouter, "detach-router", false, "Detach a router from this network.")
```

internal/commands/network/modify.go
kaminek: ok

PR #93, opened 2021-06-07
feat(server): parity between server create and network-interface create commands

aakso:
* Support all network interface related attributes during server create.
  Before this, a private interface couldn't be created as the network
  couldn't be provided.
* Use `AddEnableDisableFlags` for bootable and source ip filtering in both
  command contexts. This is to allow user to explicitly declare the intent
  regardless of the API's defaults.
* Remove check for supported families per interface type from
  `network-interface create` command. The limitation is a "soft" one and will
  probably change. It is better to let the API handle this kind of validation.
* Remove `network` context check from `network-interface create` command.
  It is supported to specify address with public and utility type of
  interfaces for instance with private clouds.

### DIFF
--- a/internal/commands/server/create.go
+++ b/internal/commands/server/create.go
@@ -6,20 +6,22 @@ import (
 	"os"
 	"strings"
 
+	"github.com/UpCloudLtd/upcloud-go-api/upcloud"
+	"github.com/UpCloudLtd/upcloud-go-api/upcloud/request"
+	"github.com/UpCloudLtd/upcloud-go-api/upcloud/service"
 	"github.com/jedib0t/go-pretty/v6/text"
+	"github.com/spf13/pflag"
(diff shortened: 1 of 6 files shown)


moitias:
LGTM :+1:

internal/commands/server/create.go
```diff
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/UpCloudLtd/upcloud-cli/internal/commands/ipaddress"
```
kaminek, line 9: put internal lib all together separated by a line with the external libs

kaminek (approved):
LGTM

internal/commands/server/create.go
aakso: changed my IDE config to use goimports with `-local` flag which seems to do the trick.

PR #98, opened 2021-09-29
Default to not create a password, but do so when password delivery method is chosen

### LINKED ISSUE #97: New OTP changes in UpCloud API results default "create_password" param to faulty server creation with SSH_keys
Expected to create a server with ssh-keys results in error:
```
upctl server create --hostname test.startup.io --ssh-keys .ssh/id_rsa.pub --zone nl-ams1
creating server test.startup.io: failed (The attribute create_password cannot be yes wh...0m0s

ERROR: The attribute create_password cannot be yes when ssh keys are provided. (CREATE_PASSWORD_INVALID)
```
This is the culprit:
https://github.com/UpCloudLtd/upcloud-cli/blob/51cf4de3eaece2cc2b8e4d78a3958d51ebbe46ba/internal/commands/server/create.go#L54

ajmyyra:
This PR will fix #97 and also address the previous insecure default of creating a password for servers. After this PR is accepted, a new version (1.1.1?) of upctl should be created.

In relation to the new [one-time password policy](https://developers.upcloud.com/1.3/8-servers/#creating-from-a-template), all deployments with both SSH keys and passwords are rejected due to the fact that one-time password would prompt for a password change for SSH logins as well, leaving the user in a nasty situation.

This PR will:
- Change the default to not creating a password
- Always creating a password when password delivery method is chosen, even if --create-password is not specified

### DIFF
--- a/internal/commands/server/create.go
+++ b/internal/commands/server/create.go
@@ -51,7 +51,7 @@ var defaultCreateParams = &createParams{
 	osStorageSize:  0,
 	sshKeys:        nil,
 	username:       "",
-	createPassword: true,
+	createPassword: false,
 }
 
 type createParams struct {
@@ -288,6 +288,10 @@ func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.
 	}
 	if s.params.os == defaultCreateParams.os && s.params.PasswordDelivery == "none" && s.params.sshKeys == nil {
 		return nil, fmt.Errorf("a password-delivery method, ssh-keys or a custom image must be specified")
+	}
+
+	if !s.createPassword.Value() && s.params.PasswordDelivery != "none" {
+		_ = s.createPassword.Set("true")
 	}
 
 	if s.params.Title == "" {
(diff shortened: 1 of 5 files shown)


internal/commands/server/create.go
```diff
@@ -290,6 +290,10 @@ func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.
 		return nil, fmt.Errorf("a password-delivery method, ssh-keys or a custom image must be specified")
 	}
 
+	if !s.createPassword.Value() && s.params.PasswordDelivery != "none" {
+		_ = s.createPassword.Set("true")
```
ka-myl, line 294: I'm not sure if I agree with this. I get the idea - if somebody sets the password delivery method then they for sure want password, but it still seems a bit too implicit to me. Not a blocker though. You might want to update the help text for `--password-delivery` flag to make sure this is documented somewhere ;)

ka-myl (approved):
LGTM

internal/commands/server/create.go
ajmyyra: As it was true by default before, this addition is for not messing with defaults for a minor version upgrade. For future versions we might think if --create-password is even needed. Maybe a --login-type would be better that would default to SSH?

Would you think this would make sense, and we should add a deprecation notice to that?

ka-myl: Switching to something like `--login-type` is imo a good idea, as it removes any ambiguity. With that it's clear that you have only one option, and we also get to clearly set the safer default 👍 
In this case, having this ☝️  behavior as a temp solution to avoid breaking change is totally ok to me :)

PR #104, opened 2022-02-01
fix: storage command attached-to-server key overrides zone

wvh:
Storage command output has a duplicate output key "zone" used for both storage Zone and attachedToServer field,
so the server ends up in the zone field and the actual zone information is overwritten.

JSON output before:

{
  ...
  "zone": "00088baf-d71c-4dd3-9b53-deadbabe"
}

JSON output after:

{
  "server": "00088baf-d71c-4dd3-9b53-deadbabe",
  "zone": "fi-hel2"
}


vtorhonen:
Thank you @wvh for the contribution! Will merge this in and it will be part of the next release (likely next week).

PR #106, opened 2022-04-04
fix(resolver): clarify error messages relating to argument resolution

moitias:
Previously, resolver failing (which can happen on backend failure, for example) would result in a stuttering error message:
```
cannot create resolver: cannot create resolver: ...
```
which also seems to indicate that it's a DNS issue which is problematic as the upctl internal use of 'resolver' refers to *argument* resolution rather than name resolution. This PR changes the error messages and the above error would look like:
```
cannot resolve command line arguments: cannot get resolver: ...
```


peknur (approved):
Looks good!

PR #110, opened 2022-04-13
Do not create API client for offline commands

### LINKED ISSUE #94: version command not working
**Environment (required):**
- OS: macOS  (Darwin)
- Version: latest as of now

**Describe the bug**
upctl version does not print the software version, see screenshot below.

**To Reproduce**
1. Download the upctl binary
2. Run command "upctl version"

**Expected behavior**
The version is printed into the terminal.

**Observed behavior**
I get an error about missing user credentials and a generic help text.

**Screenshots**
image

kangasta:
This allows running offline commands, such as upctl version, without having to define credentials.

Fixes #94

### DIFF
--- a/internal/commands/command.go
+++ b/internal/commands/command.go
@@ -30,6 +30,11 @@ type Command interface {
 	CobraCommand
 }
 
+// OfflineCommand is a command that does not need connect to the API, e.g. upctl version.
+type OfflineCommand interface {
+	DoesNotUseServices()
+}
+
 // NoArgumentCommand is a command that does not care about the positional arguments.
 type NoArgumentCommand interface {
 	Command
@@ -95,6 +100,11 @@ func BuildCommand(child Command, parent *cobra.Command, config *config.Config) C
 
 	// Set run
 	child.Cobra().RunE = func(cmd *cobra.Command, args []string) error {
+		// Do not create service for offline commands, e.g. upctl version
+		if _, ok := child.(OfflineCommand); ok {
+			return commandRunE(child, nil, config, args)
+		}
+
 		service, err := config.CreateService()
 		if err != nil {
 			return fmt.Errorf("cannot create service: %w", err)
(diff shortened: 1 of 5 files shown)


internal/commands/command.go
```diff
@@ -95,6 +100,11 @@ func BuildCommand(child Command, parent *cobra.Command, config *config.Config) C
 
 	// Set run
 	child.Cobra().RunE = func(cmd *cobra.Command, args []string) error {
+		// Do not create service for offline commands, e.g. upctl version
+		if i, ok := child.(OfflineCommand); ok && i.IsOffline() {
```
kangasta, line 104: Would there be a better way to tell `BuildCommand()` that some commands do not need services connection than this interface check?

moitias: This seems fine to me, it's not a completely foreign pattern and other options would need much more changes (lazy init of service? move the service init to `commandRunE` and make it a separate type of a command? something else?).

Although, commonly this pattern has only a member method with no return value (eg `type OfflineCommand interface { IsOffline() }`) as this prevents the confusing situation where something fulfills `OnlineCommand` but the method returns false. That is, using the fulfillment of the interface as the definition rather than what that particular method returns. If you go that route, maybe a good idea to make the name of the method a bit more explicit too - just `IsOffline` doesn't really say much unless one knows the context.

kangasta: I removed the return value and renamed `IsOffline` to `DoesNotUseServices`.

I also though of separate type of command, but that seemed like a lot of work as we probably wont have much commands that don't need the API.

moitias: That works great :+1:

moitias (approved):
One more small nit that you can choose to ignore if you prefer (sorry for not spotting it earlier) but otherwise looks great to me :+1:

internal/commands/root/version.go
```diff
@@ -17,10 +17,13 @@ type VersionCommand struct {
...
-			{Title: "Version", Key: "version", Value: config.Version},
+			{Title: "Version:", Key: "version", Value: config.Version},
 			{Title: "Build date:", Key: "build_date", Value: config.BuildDate},
 			{Title: "Built with:", Key: "built_with", Value: runtime.Version()},
 		},
 		}},
 	}, nil
 }
+
+// Implement OfflineCommand to tell BuildCommand that this command does not use services
```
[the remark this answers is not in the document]
kangasta: Done

PR #111, opened 2022-04-14
Fix/server create help

ka-myl:
This PR:

- fixes error text for using `cores` and `memory` option without plan set (the error message indicated that you can omit plan, which is not true)
- reorders flag descriptions alphabetically
- makes the `plan`, `cores` and `memory` flags descriptions a little bit more clear on how to use them
- adds example of `server create` command with flexible plan
image
(flags are ordered differently, the picture is just to show how all three options looks now)


internal/commands/server/create.go
```diff
@@ -303,11 +306,9 @@ func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.
 			return nil, fmt.Errorf("both --cores and --memory must be defined for custom plans")
 		}
 
-		if s.params.Plan != "" && s.params.Plan != customPlan {
-			return nil, fmt.Errorf("--plan needs to be 'custom' or omitted when --cores and --memory are specified")
+		if s.params.Plan != customPlan {
+			return nil, fmt.Errorf("--plan needs to be 'custom' when --cores and --memory are specified")
 		}
-
-		s.params.Plan = customPlan // Valid for all custom plans.
```
ka-myl, line 310: This is not really needed, because `Plan` has default value. So `s.params.Plan` will always be either `custom` or a simple plan string

internal/commands/server/create.go
```diff
@@ -250,34 +251,34 @@ func (s *createCommand) InitCommand() {
...
 	fs.StringVar(&s.params.BootOrder, "boot-order", def.BootOrder, "The boot device order, disk / cdrom / network or comma separated combination.")
-	fs.StringVar(&s.params.UserData, "user-data", def.UserData, "Defines URL for a server setup script, or the script body itself.")
-	fs.IntVar(&s.params.CoreNumber, "cores", def.CoreNumber, "Number of cores. Use only when defining a flexible (\"custom\") plan.")
-	fs.IntVar(&s.params.MemoryAmount, "memory", def.MemoryAmount, "Memory amount in MiB. Use only when defining a flexible (\"custom\") plan.")
-	fs.StringVar(&s.params.Title, "title", def.Title, "A short, informational description.")
+	fs.IntVar(&s.params.CoreNumber, "cores", def.CoreNumber, "Number of cores. Only allowed if `plan` option is set to \"custom\".")
+	config.AddToggleFlag(fs, &s.createPassword, "create-password", def.createPassword, "Create an admin password.")
+	config.AddEnableOrDisableFlag(fs, &s.firewall, def.firewall, "firewall", "firewall")
+	config.AddEnableOrDisableFlag(fs, &s.metadata, def.metadata, "metadata", "metadata service")
+	config.AddEnableOrDisableFlag(fs, &s.remoteAccess, def.remoteAccess, "remote-access", "remote access")
```
ka-myl, line 259: `AddEnableOrDisableFlag` prefixes flag with `enable` (so for example this particular line will be `--enable-remote-access`)

internal/commands/server/create.go
```diff
@@ -283,6 +284,8 @@ func (s *createCommand) InitCommand() {
 
 // ExecuteWithoutArguments implements commands.NoArgumentCommand
 func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Output, error) {
+	fmt.Printf("\033[31m Plan: %s | Cores: %d | Mem: %d\033[0m", s.params.Plan, s.params.CoreNumber, s.params.MemoryAmount)
```
kangasta: This now displays 0 cores and mem for simple plans, which might be confusing.

(Also this is visible for just a moment before `logline.SetMessage` is called for the first time. Could be helpful to have logging mode which would keep all messages in view instead of replacing the line.)

ka-myl: This was just a log that I used for quick check, removed the whole thing already :)

kangasta:
LGTM

kangasta (approved):

PR #112, opened 2022-04-14
Display both public and private addresses in server create output

kangasta:


internal/commands/server/create.go
```diff
@@ -372,17 +372,29 @@ func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.
...
-		for i, ipa := range ipAddresses {
-			strs[i] = ipa.Address
+	server, ok := val.(*upcloud.ServerDetails)
+	if !ok {
+		return nil, fmt.Sprint(val), nil
+	}
+
+	var strs []string
+
+	// Get puplic addressses from ip_addresses list
```
ka-myl, line 387:
```suggestion
	// Get public addressses from ip_addresses list
```

peknur (approved):
Tested this locally and didn't find any problems. Looks good to me!

peknur (approved):

PR #113, opened 2022-04-19
fix(core): disable colors on non-human output

peknur:
- disable colors on non-human output
- remove extra space after server IP
- add `.vscode` to .gitignore


ka-myl (approved):
LGTM

PR #115, opened 2022-04-19
Include UUID of created resource in create output

kangasta:
This PR also fixes livelog rendering of quick commands (execution takes less than render tick interval) and syncs livelog formatting of related commands.


peknur (approved):
Looks good!

PR #120, opened 2022-04-29
Do not show usage on missing credentials error

kangasta:
I set the target to `release/v1.2.0` assuming GitHub will automatically update that to `master` once #119 is merged.


ka-myl (approved):
Tested locally, seems to work as intended :)

PR #123, opened 2022-05-06
feat(server): add wait flag to create and stop commands

kangasta:
Script to test this:

```bash
#!/bin/bash -xe
prefix=server-wait-flag-test-

./bin/upctl server create --wait --hostname ${prefix}server --network type=public --zone pl-waw1 --ssh-keys ~/.ssh/id_ed25519.pub

# Server seems to ignore stop, if it is done right after start
sleep 15

./bin/upctl server stop --wait ${prefix}server

./bin/upctl server delete ${prefix}server
```

### DIFF
--- a/internal/commands/server/create.go
+++ b/internal/commands/server/create.go
@@ -30,6 +30,7 @@ func CreateCommand() commands.Command {
 			"create",
 			"Create a server",
 			"upctl server create --title myapp --zone fi-hel1 --hostname myapp --password-delivery email",
+			"upctl server create --wait --title myapp --zone fi-hel1 --hostname myapp --password-delivery email",
 			"upctl server create --title \"My Server\" --zone fi-hel1 --hostname myapp --password-delivery email",
 			"upctl server create --zone fi-hel1 --hostname myapp --password-delivery email --plan 2xCPU-4GB",
 			"upctl server create --zone fi-hel1 --hostname myapp --password-delivery email --plan custom --cores 2 --memory 4096",
@@ -243,6 +244,7 @@ type createCommand struct {
 	metadata       config.OptionalBoolean
 	remoteAccess   config.OptionalBoolean
 	createPassword config.OptionalBoolean
+	wait           config.OptionalBoolean
 }
 
 // InitCommand implements Command.InitCommand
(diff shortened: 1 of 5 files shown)


internal/commands/server/create.go
```diff
@@ -369,6 +372,13 @@ func (s *createCommand) ExecuteWithoutArguments(exec commands.Executor) (output.
 	logline.SetMessage(fmt.Sprintf("%s: request sent", msg))
 	logline.MarkDone()
 
+	if s.wait.Value() {
+		logline := exec.NewLogEntry(msg)
```
kangasta, line 376: Is it better to create new log line or update the existing one?

internal/commands/server/stop.go
```diff
@@ -63,5 +68,12 @@ func (s *stopCommand) Execute(exec commands.Executor, uuid string) (output.Outpu
 	logline.SetMessage(fmt.Sprintf("%s: request sent", msg))
 	logline.MarkDone()
 
+	if s.wait.Value() {
+		logline := exec.NewLogEntry(msg)
```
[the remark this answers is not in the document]
peknur: I guess it would be nice if only that one line would update because it's more of an status display rather that event logging but I don't have that strong opinion on this one.

peknur (approved):

internal/commands/server/create.go
peknur: I guess it would be nice if only that one line would update because it's more of an status display rather that event logging but I don't have that strong opinion on this one.

kangasta: I'll test how this would work if existing logline was updated instead of creating a new one.

Would it be better to log warning about failed `--wait` and details of created server, instead of only wait timeout error, if server does not reach given state within five minutes?

kangasta: I'll keep this two message approach for now. On timeout error, it is easier to see that error comes from `--wait` if the logline from main command is still visible. This is something we could revisit later, though.

Also, seems that currently it is not possible to easily output both error and output from command, so printing both timeout error and server details would require further refactoring.

peknur: Sorry for late reply. Somehow I missed your question last week :)

I think that it would better to just show warning if wait fails.

kangasta: Added new commit to use single logline and output only warning on `--wait` failure. This way the user sees the UUID and IP of created server even if `--wait` timeouts, which should be unlikely.

PR #124, opened 2022-05-06
fix(livelog): mark error and warning livelogs finished

kangasta:
This stops the timer from the end of the error and warning rows and stops livelog from refreshing lines that will not be updated.

Script to test this:
```bash
#!/bin/bash -xe
prefix=livelog-errors-test-

./bin/upctl server create --wait --hostname ${prefix}server-1 --network type=public --zone pl-waw1 --ssh-keys ~/.ssh/id_ed25519.pub
./bin/upctl server create --wait --hostname ${prefix}server-2 --network type=public --zone pl-waw1 --ssh-keys ~/.ssh/id_ed25519.pub

# Server seems to ignore stop, if it is done right after start
sleep 10

./bin/upctl server stop --wait ${prefix}server-1

# Output of this command is where the fix is visible (and bug was visible)
./bin/upctl server stop --wait ${prefix}server-1 ${prefix}server-2

./bin/upctl server delete ${prefix}server-1 ${prefix}server-2
```

Built on top of #123. Thus the target branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Do not display usage if execution fails because of missing credentials
+- Mark error and warning livelogs finished when they will not be updated anymore: this stops the timer in the end of the row and stops livelog from refreshing these lines.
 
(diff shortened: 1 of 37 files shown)


peknur (approved):
Looks good! I tested this locally and it fixes the problem.

Issue #101, opened 2021-12-09
Add Default Drop option for upctl firewall option


nicovs:
**Is your feature request related to a problem? Please describe.**
Through CLI, I need to add 2 seperate rules in the FW atm to block all trafic by default

**Describe the solution you'd like**
and option to set the default action from ACCEPT to DROP, as there is an option in the GUI to do this
```bash
upctl server firewall default-action "${SERVERNAME}" DROP
```
**Describe alternatives you've considered**
```bash
upctl server firewall create "${SERVERNAME}" --direction in --action drop --position 1000 --family IPv4 --comment "Drop all IPv4 traffic"
upctl server firewall create "${SERVERNAME}" --direction in --action drop --position 1000 --family IPv6 --comment "Drop all IPv6 traffic"
bash

kangasta:
Hello, and thank you for reporting this issue!

Creating default/catch-all rule is now possible in `v1.5.0` by only specifying `--direction` and `--action` parameters.

PR #138, opened 2022-06-22
Add --default parameter for editing default firewall rules

kangasta:
Implements #101

### TODO

- [ ] Should we check for existing default rule when trying to create new one?


internal/commands/serverfirewall/create.go
```diff
@@ -51,6 +52,11 @@ func (s *createCommand) MaximumExecutions() int {
 
 // InitCommand implements Command.InitCommand
 func (s *createCommand) InitCommand() {
+	s.Cobra().Long = `Create a new firewall rule
+
+To edit the default rule of the firewall, set only --direction and --action parameters.
+This will create catch-all rule that will take effect when no other rule matches.`
```
kangasta, line 58: Is this accurate/clear-enough?

ka-myl: Almost. The default rule should also be the last rule in the whole set. So it has to be either created last, or it has to be placed at the end using the `--position` flag

ka-myl (approved):

Issue #134, opened 2022-06-20
ZSH completion


Davincible:

kangasta:
Thank you for reporting this!

We have been planning to update our [cobra](https://github.com/spf13/cobra) dependecy to version `v1.5.0` once that is released and update our completions handling at the same time. This would likely include adding ZSH completions as well (as they are part of the completion command cobra includes by default).

So, hopefully coming soon!

kangasta:
This is now finally available in [v2.5.0](https://github.com/UpCloudLtd/upcloud-cli/releases/tag/v2.5.0)

PR #141, opened 2022-06-30
Refactor completions to use cobras default completion command

kangasta:
- [x] Depends on #139 
- [ ] Depends on fix for https://github.com/spf13/cobra/issues/1740
- [ ] Remove replace directive from go.mod

Implements #134


kangasta:
Closing this as #204 was merged in. We can handle possible cleanups in a new MR when cobra supports whitespace in bash completions.

PR #149, opened 2022-07-12
Extract progress logging into a separate module

kangasta:
Created module available in https://github.com/UpCloudLtd/progress

- [x] Rebase on top of `main`.
- [x] Update `go.mod` once there is a release available: https://github.com/UpCloudLtd/progress/pull/16


internal/commands/executor.go
```diff
@@ -2,23 +2,25 @@ package commands
...
 	"github.com/UpCloudLtd/upcloud-cli/internal/output"
 	internal "github.com/UpCloudLtd/upcloud-cli/internal/service"
-	"github.com/UpCloudLtd/upcloud-cli/internal/ui"
 
 	"github.com/UpCloudLtd/upcloud-go-api/v4/upcloud/service"
 	"github.com/gemalto/flume"
 )
 
 // Executor represents the execution context for commands
 type Executor interface {
```
villevsv-upcloud, line 18: Not totally related, but I think interface is not needed as we only seem to pass around `executorImp`. Would it make sense to remove the interface and rename the struct to `Executor` at this point?

internal/config/progress.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"os"
+
+	"github.com/UpCloudLtd/progress"
+	"github.com/UpCloudLtd/progress/messages"
+	"github.com/jedib0t/go-pretty/v6/text"
+)
+
+func GetProgressOutputConfig() *progress.OutputConfig {
```
villevsv-upcloud, line 11: Could put these settings in `messages.GetDefaultOutputConfig()`  in `progress` repo and call directly that? Or get the current default and only override the changes here.

villevsv-upcloud (approved):

internal/commands/util.go
```diff
@@ -129,10 +129,12 @@ func BoolFromString(b string) (*upcloud.Boolean, error) {
 }
 
 // HandleError logs error in livelog by setting message to msg and details to err. Returns (nil, err), where err is the err passed in as input.
-func HandleError(logline *ui.LogEntry, msg string, err error) (output.Output, error) {
-	logline.SetMessage(msg)
-	logline.SetDetails(err.Error(), "Error: ")
-	logline.MarkFailed()
+func HandleError(exec Executor, key string, err error) (output.Output, error) {
```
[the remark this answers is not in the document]
kangasta: Good idea, done!

internal/commands/executor.go
kangasta: I think we could create a separate ticket for that as this is already quite big PR

internal/config/progress.go
kangasta: Done. This now uses default config as base and colors whole message with bright colors.

PR #150, opened 2022-07-13
Disable colors if `NO_COLOR` env variable is set

kangasta:


aakso:
Perhaps rename then flag to NO_COLOURS for consistency with the flag? Of course we could have both spellings.

kangasta:
Many other CLI apps use [NO_COLOR](https://no-color.org/) so for compatibility it would be better even though it looks a bit odd that we use US spelling for env variable but not in options.

kangasta:
But true, we might as well check for both spellings to be both consistent and compatible

ka-myl (approved):
LGTM.

PR #155, opened 2022-07-25
Update LICENSE

rawidea:


kangasta:
Thank you for the interest towards this project and sorry for taking this long to respond!

We discussed this internally and, instead of bumping the year, decided to use year range. This is to include both initial release year and years with major changes in the copyright notice. This is done in #167. Thus, I'll close this PR.

PR #161, opened 2022-08-09
feat(database): delete command

villevsv-upcloud:


internal/commands/database/delete_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			mService := smock.Service{}
+			mService.On(targetMethod, &test.req).Return(nil)
+
+			conf := config.New()
+			c := commands.BuildCommand(DeleteCommand(), nil, conf)
+
+			_, err := c.(commands.MultipleArgumentCommand).Execute(commands.NewExecutor(conf, &mService, flume.New("test")), test.arg)
```
kangasta, line 43: Alternatively, [mockexecute](https://github.com/UpCloudLtd/upcloud-cli/blob/main/internal/mockexecute/mockexecute.go) helper could be used to simulate the command execution, but as it mostly helps testing Cobras argument validations and checking output, it would not bring much benefits here.

kangasta:
PR could also include a CHANGELOG entry. Otherwise, looks good to me!

kangasta (approved):

PR #164, opened 2022-08-09
ci: golangci-lint@v3 and go-version 1.18 pinned

villevsv-upcloud:


.golangci.yaml
```diff
@@ -37,12 +41,23 @@ linters:
...
+
 linters-settings:
-  gocritic:
-    disabled-checks:
-      # TODO: re-enable after figuring out why runcommand fails on go-critic..
-      - unlambda
   goconst:
     min-len: 5
+  predeclared:
+    ignore: "new"
```
kangasta, line 63: I added this ignore to Terraform provider as it had many instances of `old, new := d.GetChange(...)`, probably would have been good to add explanation there as well 😅 If `upctl` does not have `new` declarations at the moment this could be not ignored.

kangasta (approved):

kangasta:
Looks good to me! Didn't do any manul e2e testing as this should be a low risk change.

kangasta (approved):

PR #165, opened 2022-08-10
Fix `network show` output for networks that contains a load balancer

kangasta:
To be rebased on top of linter fixes:
- #162 
- #164

### DIFF
--- a/internal/commands/util.go
+++ b/internal/commands/util.go
@@ -96,34 +96,6 @@ func LoadBalancerOperationalStateColour(state upcloud.LoadBalancerOperationalSta
 	}
 }
 
-// ServerStateColour is a helper mapping server states to colours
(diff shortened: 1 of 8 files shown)


internal/commands/util.go
```diff
@@ -110,6 +110,20 @@ func ServerStateColour(state string) text.Colors {
 	}
 }
 
+// FormatServerState implements Format function for server states
```
kangasta, line 113: Is this good place for this format function or should it be in `ui` package, similarly than `FormatPossiblyUnknownString`.

kangasta: Created new format package for these. We could later also move/add LB and DB state formatters there.

villevsv-upcloud (approved):

PR #169, opened 2022-08-26
refactor(format): use Format functions to color resource states

kangasta:


internal/format/common.go
```diff
@@ -18,3 +18,20 @@ func PossiblyUnknownString(val interface{}) (text.Colors, string, error) {
 	}
 	return nil, str, nil
 }
+
+func usingColorFunction[T ~string](colorFunction func(T) text.Colors) func(interface{}) (text.Colors, string, error) {
```
kangasta, line 22: Possibly first generics use-case in this project!

peknur: I would keep it simple with something like this:
```
func usingColorFunction(colorFunction func(string) text.Colors, val string) (text.Colors, string, error) {
	if val == "" {
		return PossiblyUnknownString(val)
	}
	return nil, colorFunction(val).Sprint(val), nil
}
````
then do something like this e.g. in database color function:
```
func databaseStateColour(state string) text.Colors {
	switch upcloud.ManagedDatabaseState(state) { 
	...
```
and do type assertion (and possible error checking) in state function:
```
func DatabaseState(val interface{}) (text.Colors, string, error) {
	return usingColorFunction(databaseStateColour, val.(string))
}
```
I guess you could do type assertion also in `usingColorFunction` and check errors there but for me less empty interfaces I see is always win :)

peknur: and to answer to you original question I would say that without currying.

peknur (approved):
Looks good! Left some ideas in comments but no blockers.

internal/format/common.go
kangasta: I removed the currying. The function looks nicer now.

To me it felt clearer to have both type check and empty value check in same place and, as those are almost the same in all cases, helper function seemed like good choice.

Issue #174, opened 2022-09-02
Please add a subcommand to `upctl storage` to wait for a given storage to be online


nogweii:
**Is your feature request related to a problem? Please describe.**
I'm trying to automate the process of uploading a raw Arch Linux disk image (from their [arch-boxes](https://gitlab.archlinux.org/archlinux/arch-boxes) project) and each step needs the storage to be online. I had to write my own loop around that, and would appreciate the CLI providing that power for me.

**Describe the solution you'd like**
An additional subcommand called "wait" or similar that can take a UUID (and/or title?) and will block until the storage resource is in an online state, ready to be used.

**Describe alternatives you've considered**
I have to write a `while` loop in each of my scripts wrapping `upctl`. Not that I have many, though.

**Additional context**
I built some automation around `upctl` to do what I wanted, but it would be nice if that could have been easier. You can see a copy of [my code in this gist](https://gist.github.com/nogweii/0d9afa8246525d21eaf71ad2d12ae955);

Specifically, I'd like to call out the following lines:

https://gist.github.com/nogweii/0d9afa8246525d21eaf71ad2d12ae955#file-upload_to_upcloud-sh-L22-L29

kangasta:
Hi, thanks for the improvement idea!

We currently have similar behavior for server create and stop commands implemented with `--wait` flag, e.g. `upctl server stop --wait $server-name`.

Following the same pattern, we could add `--wait` flag to `upctl storage import` as well. This would work at least in the gist you provided, right?

nogweii:
Yep, along with the same flag being available for `upctl storage templatise` as well. Add the flag to both would satisfy my use case.

PR #175, opened 2022-09-04
 feat(storage): add `--wait` flag to import and templatise commands 

kangasta:
Implements #174 and in addition:
- Add target storage info to machine readable `storage import` output
- Display UUID of created storage, if one was created, in human output of `storage import` command

### DIFF
--- a/internal/commands/storage/import.go
+++ b/internal/commands/storage/import.go
@@ -84,14 +84,30 @@ type storageImportStatus struct {
 	complete bool
 }
 
-func getImportSuccessOutput(res *upcloud.StorageImportDetails) (output.Output, error) {
-	return output.MarshaledWithHumanDetails{
-		Value: res,
-		Details: []output.DetailRow{{
-			Title:  "UUID",
-			Value:  res.UUID,
-			Colour: ui.DefaultUUUIDColours,
-		}},
+type storageImportOutput struct {
+	*upcloud.StorageImportDetails
+	Storage upcloud.Storage `json:"storage"`
+}
+
+func getImportSuccessOutput(res *upcloud.StorageImportDetails, storage upcloud.Storage, isNewStorage bool) (output.Output, error) {
+	value := storageImportOutput{
+		StorageImportDetails: res,
+		Storage:              storage,
+	}
+
+	if isNewStorage {
+		return output.MarshaledWithHumanDetails{
+			Value: value,
+			Details: []output.DetailRow{{
+				Title:  "UUID",
+				Value:  storage.UUID,
+				Colour: ui.DefaultUUUIDColours,
+			}},
+		}, nil
+	}
+
+	return output.OnlyMarshaled{
+		Value: value,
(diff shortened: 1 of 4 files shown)


kangasta:
Example commands to test this with:

```sh
upctl storage import --source-location https://cdimage.ubuntu.com/jammy/daily-live/current/jammy-desktop-amd64.iso --title jammy-desktop-amd64-daily --zone pl-waw1 --wait

upctl storage templatise jammy-desktop-amd64-daily --title jammy-daily-cd --wait
```

internal/commands/storage/import.go
```diff
@@ -69,6 +70,7 @@ func (s *importCommand) InitCommand() {
 	flagSet.StringVar(&s.sourceLocation, "source-location", "", "Location of the source of the import. Can be a file or a URL.")
 	flagSet.StringVar(&s.existingStorageUUIDOrName, "storage", "", "Import to an existing storage. Storage must be large enough and must be undetached or the server where the storage is attached must be in shutdown state.")
 	config.AddToggleFlag(flagSet, &s.noWait, "no-wait", false, "Do not wait until the import finishes. Only applicable when importing from a remote URL.")
+	config.AddToggleFlag(flagSet, &s.wait, "wait", false, "Wait for storage to be in online state before returning.")
```
kangasta, line 73: If both `--no-wait` and `--wait` are used, `--no-wait` takes precedence as that code branch does not check for `--wait` value. Didn't add explanation to help texts as I assumed that these are not likely used together 😅

peknur: This is bit confusing. My first thought would be that If `--no-wait` is not set then we should wait.

kangasta: Yeah, currently `--no-wait` returns after url-to-cloud import is started and does not affect other import types. So it is has different scope than this new `--wait` would have.

Renaming the `--wait` parameter to `--wait-online` would make this bit clearer without being a breaking change. However, then we would have `--wait` parameter for async-by-default server operations and `--wait-online` paramenter for async-by-default storage operations.

Not sure what would be the best approach here.

peknur: Yeah I guess we can't have perfect solution without doing breaking change but I would vote for changing it to `--wait-online`  and then sort these naming conflicts out in v3.

villevsv-upcloud: I would vote for keeping the new flag as `--wait` and somehow explain in the help how these two are (not) related. Leaning towards this to as the server's `--wait` more or less already acts in a similar fashion from the user point of view.

In that case, we could rename `--no-wait` to something more import-related in the next major release.

peknur (approved):
Tested locally and works as expected.

internal/commands/storage/import.go
kangasta: I tried to modify `--no-wait` description to make the difference more clear, and created a ticket for renaming it in next major release.

Issue #177, opened 2022-09-09
CLI prints to stderr 


Davincible:
For some reason the CLI now prints to stderr instead of to stdout, making piping more difficult.

```bash
# Empty output
upctl server list 2>/dev/null

# Prints everything
upctl server list 1>/dev/null
```

```
Version:    2.1.0                
Build date: 2022-09-07T10:46:33Z 
Built with: go1.18.5
```

Davincible:
I now see this is listed in the release notes

`Breaking: Human output, including errors, is written to stderr instead of stdout.`

... why?

kangasta:
Hi, sorry for the confusion!

The idea was that human output is not ideal for passing to a another command for processing, but of course there are some other use cases as well, such as piping the output to `grep`. We are currently refactoring the output logic and could further tune how the output streams are used for human output.

What is your use-case for piping? I'm thinking if there is something we could add something on the `upctl` side to remove the need for piping.

Davincible:
I pipe to e.g. check if servers are up and running, as a step in a bigger script. 
Fixed it now with `-ojson | jq`.

I would say that human and machine output can both go to stdin since you only have one at a time. If you specify machine output you won't have human output anyways.

kangasta:
This will soon be fixed in `v2.2.0`

PR #180, opened 2022-09-29
fix(output): output normal human output to stdout

kangasta:
This fixes issue described in #177 and also removes duplicate logging of errors when errors are logged to progress log.


internal/output/error.go
```diff
@@ -26,15 +25,8 @@ func (e Error) MarshalJSON() ([]byte, error) {
...
-	if e.Value == nil {
-		return []byte("\nERROR: Unspecified error"), nil
-	}
-	s := fmt.Sprintf("\nERROR: %v", e.Value.Error())
-	if e.Original != "" {
-		s += fmt.Sprintf(" (%s)", e.Original)
-	}
-
-	return []byte(s), nil
+	// Human readable errors go to stderr. This output package is responsible for stdout output only.
```
kangasta, line 28: This has not been the case earlier, but in my opinion limiting this `output` package to handle only `stdout` would make sense.

internal/commands/error.go
```diff
@@ -13,5 +13,41 @@ func HandleError(exec Executor, key string, err error) (output.Output, error) {
 		Details: "Error: " + err.Error(),
 	})
 
-	return nil, err
+	return nil, handledError{err}
+}
+
+type iHandledError interface {
+	HasBeenLogged()
+}
```
peknur, line 21: Do we need this interface if we just check `err.(handledError)` in `outputError` function ?

internal/output/error.go
```diff
@@ -26,15 +25,8 @@ func (e Error) MarshalJSON() ([]byte, error) {
...
+	return []byte(""), nil
```
peknur, line 29: What if we would return `return nil, errors.New("output package is responsible for stdout output only")` (or some error) here so that we don't expect this function returning something sensible in other parts of code?

peknur (approved):
Looks good to me. Left some comments but no blockers.

internal/commands/error.go
kangasta: Good point! Removed this extra interface.

internal/output/error.go
kangasta: I think that would require bigger refactor as these `Marshal*` functions are currently called for every commands `Output` without checking the value before call and any error will stop output rendering. This is now special case as we include errors in machine readable normal output (stdout) but omit them from human normal output.

PR #188, opened 2022-10-24
refactor(docs): generate documentation with Material for MkDocs

kangasta:
Remove previous Jekyll based GitHub pages and replace those with MkDocs based solution that uses Material for MkDocs theme with few custom overrides. Remove generated documentation from `main` branch.

In addition, add custom logic for markdown command reference generation in order to:
- list command aliases in markdown files,
- present command options in tables for increased accessibility,
- display only long or short command description similarly than in `--help` outputs,
- and use `*cobra.Command` instead of output filename as input to file prefix function.


kangasta:
To test this:

```sh
git fetch
git checkout gh-pages
python3 -m http.server 8000 # or any other command that starts static HTTP server
```

After merge:
- Update GitHub pages settings to use `gh-pages` branch as is instead of Jekyll from `main`
- Add link to GitHub pages to repository website setting

villevsv-upcloud (approved):
Clean approach and easy to review.

Dead link to `docs/upctl.md` in `README.md` would need to be changed, otherwise LGTM.

kangasta:
Good catch! I'll update Documentation section in README before merging this

peknur (approved):
Good looking docs!

PR #192, opened 2022-11-04
feat: parse version from BuildInfo if it was not defined during build

kangasta:
This adds meaningful version info also to binaries created with `go install` without defining `-ldflags` to override `.../config.Version` variable.


peknur (approved):
Looks good! Tested locally and got expected `v2.2.0-9-gc95605f`

villevsv-upcloud:
> Looks good! Tested locally and got expected `v2.2.0-9-gc95605f`

Same & same!

villevsv-upcloud (approved):

PR #193, opened 2022-11-09
feat(containers): use alpine as base to have shell in our images

kangasta:
This makes the `upcloud/upctl` container more suitable for usage in CI pipelines. This also adds ~5 MB (size of `alpine:3.16` vs. `scratch`) to the image size and thus the image size will be ~16 MB in total.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,7 +7,7 @@ before:
 builds:
   - env:
       - CGO_ENABLED=0
-      - GO_MODULE_NAME=github.com/UpCloudLtd/upcloud-cli
+      - GO_MODULE_NAME=github.com/UpCloudLtd/upcloud-cli/v2
     goos:
       - linux
       - windows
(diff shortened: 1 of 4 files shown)


kangasta:
Not sure what would be the best approach for testing the goreleaser logic with the new goreleaser Dockerfile. My laptop seems to have incompatible docker and goreleaser installations, so I had to comment out buildx when testing the container image build locally.

peknur:
> Not sure what would be the best approach for testing the goreleaser logic with the new goreleaser Dockerfile. My laptop seems to have incompatible docker and goreleaser installations, so I had to comment out buildx when testing the container image build locally.

Can we have two builds e.g. id `buildx` and `docker` ? Something like:
```yaml
dockers:
  - id: buildx
    use: buildx
    # ..
  - id: docker
    use: docker
    skip_push: true
    # ..
```
In that case, we might need also to customize `release` section also to only release id `buildx`

kangasta:
Tried now in Ubuntu 22.04 VM and the build worked fine with current configuration as there buildx plugin is included in the docker installation by default. Found a bug (missing `/v2` suffix) in goreleaser configuration introduced by #191 and added commit to fix that.

peknur (approved):
LGTM

PR #195, opened 2022-11-22
feat(kubernetes): config

villevsv-upcloud:
Add `kubernetes config` command to be printed via stdout or written in a given file.


internal/commands/all/all.go
```diff
@@ -127,6 +127,7 @@ func BuildCommands(rootCmd *cobra.Command, conf *config.Config) {
 	// Kubernetes
 	kubernetesCommand := commands.BuildCommand(kubernetes.BaseKubernetesCommand(), rootCmd, conf)
 	commands.BuildCommand(kubernetes.CreateCommand(), kubernetesCommand.Cobra(), conf)
+	commands.BuildCommand(kubernetes.ConfigCommand(), kubernetesCommand.Cobra(), conf)
```
kangasta, line 130: Would it make sense to have this as `kubeconfig show` and have additional `kubeconfig add` and `kubeconfig remove` commands to add and remove clusters from users default kubeconfig (`~/.kube/config`)?

villevsv-upcloud: We could think about that for the next release! Good idea.

peknur (approved):
Looks good! Tested this locally and all worked as expected.

PR #198, opened 2022-12-07
test(config): username and password check

villevsv-upcloud:
Changed config test to check if any username and any password are set as environmental values conflicted with the original version.


peknur (approved):
Looks good to me.

PR #203, opened 2023-01-27
feat(kubernetes): add commands for nodegroup actions

kangasta:


internal/commands/kubernetes/nodegroup/scale.go
```diff
@@ -0,0 +1,61 @@
...
+
+type scaleCommand struct {
+	*commands.BaseCommand
+	name  string
+	count int
+	completion.Kubernetes
+	resolver.CachingKubernetes
+}
+
+// ScaleCommand creates the "kubernetes nodegroup delete" command
```
peknur, line 22: "delete" should probably be "scale".

internal/commands/kubernetes/nodegroup/create.go
```diff
@@ -0,0 +1,179 @@
...
+	split := strings.SplitN(in, "=", 2)
+	if len(split) < 2 {
+		return upcloud.Label{}, fmt.Errorf("invalid label: %s", in)
+	}
+
+	return upcloud.Label{
+		Key:   split[0],
+		Value: split[1],
+	}, nil
+}
```
peknur, line 65: Would it be good to support also empty values using something like `upctl ... --label=demo --label=env=dev` ?

Just as a alternative approach we could use `url.ParseQuery` here e.g.
```go
func processLabel(in string) (upcloud.Label, error) {
	q, err := url.ParseQuery(in)
	if err != nil {
		return upcloud.Label{}, err
	}
	for k := range q {
		return upcloud.Label{
			Key:   k,
			Value: q.Get(k),
		}, nil
	}
	return upcloud.Label{}, fmt.Errorf("unable to create label from input '%s'", in)
}
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - Print warning about unknown resource state before exiting when execution is interrupted with SIGINT.
+- Add `kubernetes nodegroup create`, `kubernetes nodegroup scale`, and `kubernetes nodegroup delete` commands (EXPERIMENTAL)
```
villevsv-upcloud, line 11: I was thinking that we could mention here that creating a node group with `kubernetes create` is still possible, but maybe it's too verbose for the change log. Up to the author :)

villevsv-upcloud (approved):

internal/commands/kubernetes/nodegroup/create.go
kangasta: Good idea! I'll edit the function and add test case for empty label value

kangasta: Added cases for `len(split)` instead of using `url.ParseQuery` as values can contain `&` characters that might be problematic when parsing query string if input is not url encoded first

PR #204, opened 2023-02-01
feat(completion): use cobras default completion command

kangasta:
This requires removing whitespace from completions as cobras bash completions do not support values with whitespace (See https://github.com/spf13/cobra/issues/1740). This is done by replacing whitespace with non-breaking spaces during completion generation. The argument resolvers with values that can include whitespace are also updated to handle both modified and original arguments, i.e. user can still input whitespaces normally when argument is quoted or escaped.


kangasta:
I know this is quite hacky, but so is our current bash completion override which prevents us from supporting zsh (#134) and other shell completions.

internal/completion/helpers.go
```diff
@@ -1,30 +1,28 @@
...
 func MatchStringPrefix(vals []string, key string, caseSensitive bool) []string {
 	var r []string
 	key = strings.Trim(key, "'\"")
-
 	for _, v := range vals {
 		if (caseSensitive && strings.HasPrefix(v, key)) ||
 			(!caseSensitive && strings.HasPrefix(strings.ToLower(v), strings.ToLower(key))) ||
 			key == "" {
-			r = append(r, Escape(v))
+			r = append(r, RemoveWordBreaks(v))
```
peknur, line 24: Maybe compile that regular expression before loop and just do `re.ReplaceAllString` in loop.

peknur (approved):
Looks good. Left non-blocking comment about RE.

internal/completion/helpers.go
kangasta: Good idea! Moved the re definition to package-level var.

kangasta:
The linting error seems to come from generics usage in node group parameters parsing, I'll try to see if there a route around that

PR #206, opened 2023-02-06
fix(database): parse version from metadata instead of properties

kangasta:


internal/commands/database/show.go
```diff
@@ -128,6 +128,18 @@ func (s *showCommand) Execute(exec commands.Executor, uuid string) (output.Outpu
 	}, nil
 }
 
+func getVersion(db *upcloud.ManagedDatabase) string {
+	switch db.Type {
```
peknur, line 132: Might be good to check if `db.Metadata` is `nil` ?

kangasta: Added nil check and test for this helper

peknur (approved):

PR #210, opened 2023-02-20
feat: add first completions for command arguments

kangasta:


internal/commands/kubernetes/create.go
```diff
@@ -84,7 +93,6 @@ func processNodeGroup(in string) (upcloud.KubernetesNodeGroup, error) {
 type createCommand struct {
 	*commands.BaseCommand
 	params createParams
-	completion.Kubernetes
```
kangasta, line 87: Do not suggest existing clusters as positional argument

peknur (approved):
Looks good :+1:  Tested this locally and it worked nicely.

PR #223, opened 2023-05-31
feat(dbaas): OpenSearch service type support

villevsv-upcloud:


internal/commands/database/index/list.go
```diff
@@ -0,0 +1,69 @@
...
+		rows = append(rows, output.TableRow{
+			index.IndexName,
+			index.CreateTime,
+			index.Health,
+			index.Status,
+			index.Docs,
+			index.NumberOfReplicas,
+			index.NumberOfShards,
+			index.ReadOnlyAllowDelete,
+			fmt.Sprintf("%d bytes", index.Size),
```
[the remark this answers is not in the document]
villevsv-upcloud: Couldn't find one but indeed a good idea to add it later. Task added.

internal/commands/all/all.go
```diff
@@ -134,6 +135,11 @@ func BuildCommands(rootCmd *cobra.Command, conf *config.Config) {
 		commands.BuildCommand(databaseproperties.ShowCommand(i.serviceType, i.serviceName), typeCommand.Cobra(), conf)
 	}
```
kangasta, line 137: Could also add OpenSearch as subcommand into above properties command, if it has properties available

internal/commands/database/index/delete.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type deleteCommand struct {
+	*commands.BaseCommand
+	name string
+	completion.Database
+	resolver.CachingDatabase
+}
+
+// DeleteCommand creates the "kubernetes indices delete" command
```
kangasta, line 22: `kubernetes indices` → `database index`

kangasta:
Added comment about `database properties` (and outdated code-comment), otherwise LGTM!

internal/commands/all/all.go
villevsv-upcloud: Added

internal/commands/database/index/delete.go
villevsv-upcloud: Fixed!

PR #224, opened 2023-06-06
feat: add common description and completions for `--zone` arguments

kangasta:


peknur (approved):
Good improvement :+1:

PR #227, opened 2023-06-08
feat: `servergroup`: `create`, `delete`, `list`, `modify`, `show`

villevsv-upcloud:


internal/commands/servergroup/list.go
```diff
@@ -0,0 +1,47 @@
...
+	rows := []output.TableRow{}
+	for _, serverGroup := range clusters {
+		rows = append(rows, output.TableRow{
+			serverGroup.UUID,
+			serverGroup.Title,
+			serverGroup.AntiAffinityPolicy,
+		})
+	}
+
+	return output.Table{
```
kangasta, line 39: Hmm, would the JSON and YAML pass-through work here as well (similarly than with `show` command)? 🤔 We don't use that in other lists either, but script usage might benefit from those omitted fields.

internal/commands/servergroup/list.go
```diff
@@ -0,0 +1,47 @@
...
+		Columns: []output.TableColumn{
+			{Key: "uuid", Header: "UUID", Colour: ui.DefaultUUUIDColours},
+			{Key: "title", Header: "Title"},
+			{Key: "anti_affinity", Header: "Anti-affinity policy"},
+		},
```
kangasta, line 44: Should we also display number of servers for each group? Hub also displays summary of AA status, which might be helpful as well.

internal/commands/servergroup/list.go
villevsv-upcloud: Thanks for the suggestion, now implemented! I'll make a note to add the pass-through to other resource lists as well (where applicable).

internal/commands/servergroup/list.go
villevsv-upcloud: Both implemented, good improvement indeed

internal/commands/servergroup/show.go
```diff
@@ -0,0 +1,106 @@
...
+								{Title: "Title:", Value: serverGroup.Title},
+								{Title: "Anti-affinity policy:", Value: serverGroup.AntiAffinityPolicy},
+							},
+						},
+					},
+				},
+			},
+			labels.GetLabelsSection(serverGroup.Labels),
+			output.CombinedSection{
+				Contents: antiAffinityStatus,
```
kangasta, line 102: Could we use similar table here as in server list output, but with plan and status changed into host and AA status? For example:

```txt
UUID                                   Hostname        Zone        Host         Anti-Affinity state 
────────────────────────────────────── ─────────────── ─────────── ──────────── ────────────────────
 007abd4b-ead7-4160-976c-fd1593b44c7b   orange-mk2pk    de-fra1     1234567890   unmet 
 00f5cd98-76c7-4404-8fee-fe4622901d03   orange-nfktc    de-fra1     1234567890   unmet
```

internal/commands/servergroup/modify.go
```diff
@@ -0,0 +1,109 @@
...
+	return nil
+}
+
+// InitCommand implements Command.InitCommand
+func (s *modifyCommand) InitCommand() {
+	fs := &pflag.FlagSet{}
+	fs.StringVar(&s.params.antiAffinityPolicy, "anti-affinity-policy", defaultModifyParams.antiAffinityPolicy, "Anti-affinity policy. Valid values are `yes` (best effort), `strict` and `no`. Will take effect upon server start.")
+	fs.StringArrayVar(&s.params.labels, "label", defaultModifyParams.labels, "Labels to describe the server in `key=value` format, multiple can be declared. If set, all the existing labels will be replaced with provided ones.\nUsage: --label env=dev\n\n--label owner=operations")
+	fs.StringVar(&s.params.Title, "title", defaultModifyParams.Title, "New server group title.")
+	fs.StringArrayVar(&s.params.servers, "server", defaultModifyParams.servers, "Servers that belong to the server group, multiple can be declared. If set, all the existing server entries will be replaced with provided ones.\nUsage: --server my-server\n\n--server 00333d1b-3a4a-4b75-820a-4a56d70395dd")
```
kangasta, line 84: Could have completion for `--server` argument similarly than there is in `servergroup create` command

internal/commands/servergroup/show.go
villevsv-upcloud: Using table now, looks better. Added the combined state and server count to overview as well.

internal/commands/servergroup/modify.go
villevsv-upcloud: Added

internal/commands/servergroup/show.go
```diff
@@ -0,0 +1,104 @@
...
+			serverDetails.Hostname,
+			serverDetails.Zone,
+			strconv.Itoa(serverDetails.Host),
+			string(status.Status),
+		})
+	}
+
+	serverColumns := []output.TableColumn{
+		{Key: "uuid", Header: "UUID", Colour: ui.DefaultUUUIDColours},
+		{Key: "hostname", Header: "Hostname:"},
```
kangasta, line 66: Extra trailing `:` chars in the titles that we don't seem to use in other tables

kangasta (commented):
Few extra trailing `:`s in table headers, otherwise LGTM!

internal/commands/servergroup/show.go
villevsv-upcloud: Fixed

internal/commands/servergroup/list.go
```diff
@@ -0,0 +1,62 @@
...
+	svc := exec.All()
+	serverGroups, err := svc.GetServerGroups(exec.Context(), &request.GetServerGroupsRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	rows := []output.TableRow{}
+	for _, serverGroup := range serverGroups {
+		statusSummary := "-"
+		for _, status := range serverGroup.AntiAffinityStatus {
```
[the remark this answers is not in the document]
villevsv-upcloud: Status handling fixed a bit. Using `-` as an empty placeholder but it's easy to change via constant.

internal/commands/servergroup/show.go
```diff
@@ -0,0 +1,126 @@
...
+
+			if groupStatus != unMet {
+				groupStatus = status
+			}
+		}
+	}
+
+	serverRows := []output.TableRow{}
+	for _, serverUUID := range serverGroup.Members {
+		serverDetails, err := svc.GetServerDetails(exec.Context(), &request.GetServerDetailsRequest{UUID: serverUUID})
```
kangasta, line 60: I think in some commands (at least `network show`) we leave out details from the related row, if getting details for some resource fails. Not sure if it's possible for subaccount to see that server is part of server group, but not have access to that server.

kangasta: Tested this now and seems that subaccount only has access to servergroup if it has permissions to all of its servers. So this should not cause any issues 👌

kangasta (approved):

internal/commands/servergroup/show.go
villevsv-upcloud: Excellent, thanks for the comments and testing!

PR #229, opened 2023-06-14
feat(server): label flag to create and modify commands

villevsv-upcloud:


internal/commands/server/modify.go
```diff
@@ -54,6 +56,7 @@ func (s *modifyCommand) InitCommand() {
 	flags.StringVar(&s.params.Hostname, "hostname", defaultModifyParams.Hostname, "Hostname.")
 	config.AddEnableDisableFlags(flags, &s.firewall, "firewall", "firewall")
 	// flags.StringVar(&s.params.Firewall, "firewall", defaultModifyParams.Firewall, "Enables or disables firewall on the server. You can manage firewall rules with the firewall command.\nAvailable: true, false")
+	flags.StringArrayVar(&s.params.labels, "label", defaultModifyParams.labels, "Labels to describe the server in `key=value` format, multiple can be declared.\nUsage: --label env=dev\n\n--label owner=operations")
```
peknur, line 59: I guess we could mention here that if label(s) are provided, existing labels are replaced with new labels.

peknur (approved):
Looks good :+1:  Did some testing and didn't spot any problems.

internal/commands/server/modify.go
villevsv-upcloud: Good point, added!

PR #230, opened 2023-06-15
Add `go install` to README

thevilledev:
Easy to miss and install an older `v1` (as indexed in [Go pkgs here](https://pkg.go.dev/github.com/UpCloudLtd/upcloud-cli)) version otherwise.


peknur (approved):
:+1:

PR #231, opened 2023-06-15
chore: use Go 1.20

thevilledev:
- Use Go 1.20
- Read Go version in `setup-go` CI step from `go.mod`
- Remove deprecated linters
- Increase timeout for linter since it seems to hit the default every now and then


kangasta:
Could update other pipelines as well, those would still be 1.18?

kangasta (approved):
LGTM, PR title does not follow conventional commits, but that can be tuned when merging

PR #236, opened 2023-06-28
feat(kubernetes): `--disable-utility-network-access` to nodegroup

villevsv-upcloud:


internal/commands/kubernetes/nodegroup/create.go
```diff
@@ -38,6 +40,7 @@ func GetCreateNodeGroupFlagSet(p *CreateNodeGroupParams) *pflag.FlagSet {
 	fs.StringArrayVar(&p.SSHKeys, "ssh-key", []string{}, "SSH keys to be configured as authorized keys to the nodes.")
 	fs.StringVar(&p.Storage, "storage", "", "Storage template to use when creating the nodes. Defaults to `UpCloud K8s` public template.")
 	fs.StringArrayVar(&p.Taints, "taint", []string{}, "Taints to be configured to the nodes in `key=value:effect` format")
+	config.AddEnableOrDisableFlag(fs, &p.UtilityNetworkAccess, true, "utility-network-access", "utility network access. Enabled by default")
```
kangasta, line 43: Help text could be a bit more descriptive, e.g. `utility network access. Enabled by default` → `Connect worker nodes to the utility network. Enabled by default.` would be more similar to `k8s create` commands `--private-node-groups`

kangasta (approved):

internal/commands/kubernetes/nodegroup/create.go
kangasta: Or `If set to false, nodes in this group will not have access to utility network.` from the TF provider would also be good 👌

villevsv-upcloud: Thanks! Using now the TF provider comment style.

PR #238, opened 2023-07-06
chore: prepare release v2.9.1

villevsv-upcloud:
Prepares the release & updates goreleaser config to latest schema.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -157,3 +157,48 @@ dockers:
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - "--label=org.opencontainers.image.version={{.Version}}"
       - "--label=org.opencontainers.image.source=https://github.com/UpCloudLtd/upcloud-cli.git"
+
+release:
+  # Repo in which the release will be created.
+  # Default is extracted from the origin remote URL or empty if its private hosted.
+  github:
+    owner: UpCloudLtd
+    name: upcloud-cli
+
+  # If set to true, will not auto-publish the release.
+  # Available only for GitHub and Gitea.
+  #
+  # Default: false
+  draft: true
+
+  # Whether to remove existing draft releases with the same name before creating
+  # a new one.
+  # Only effective if `draft` is set to true.
+  # Available only for GitHub.
+  #
+  # Default: false
+  # Since: v1.11
+  replace_existing_draft: true
+
+  # If set to auto, will mark the release as not ready for production
+  # in case there is an indicator for this in the tag e.g. v1.0.0-rc1
+  # If set to true, will mark the release as not ready for production.
+  # Default is false.
+  prerelease: false
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.9.1] - 2023-07-06
```
kangasta, line 10: We could also use 2.9.0 since there is no release for that and no changes to application code after that, but might bit of a bad practice to move that git tag forward

RELEASING.md
```diff
@@ -4,5 +4,10 @@
 2. Test GoReleaser config with `goreleaser check`
 3. Tag a commit with the version you want to release e.g. `v1.2.3`
 4. Push the tag & commit to GitHub
-    - GitHub actions will automatically set the version based on the tag, create a GitHub release, build the project, and upload binaries & SHA sum to the GitHub release
-5. [Edit the new release in GitHub](https://github.com/UpCloudLtd/upcloud-cli/releases) and add the changelog for this release
\ No newline at end of file
+    - GitHub action automatically
+      - sets the version based on the tag
+      - creates a draft release to GitHub
+      - populates the release notes from `CHANGELOG.md` with `make release-notes`
```
kangasta, line 10: Nice 👌

.goreleaser.yaml
```diff
@@ -42,127 +49,59 @@ snapshot:
 
 changelog:
```
kangasta, line 50: Is this needed if we use `--changelog` with `make release-notes` output?

.goreleaser.yaml
```diff
@@ -42,127 +49,59 @@ snapshot:
...
-
-    # Git author used to commit to the repository.
-    # Defaults are shown.
     commit_author:
       name: goreleaserbot
       email: goreleaser@carlosbecker.com
 
-    # Folder inside the repository to put the formula.
-    # Default is the root folder.
-    folder: Formula
```
kangasta, line 154: Homebrew seems to [recommend](https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap#creating-a-tap) using Formulas folder, but both should work.

.goreleaser.yaml
villevsv-upcloud: Not needed, but I suppose it's better to have it for the sake of keeping `skip: false` in clear sight. Just added the skip to this section with the default comment so we know not to skip the changelog here either.

.goreleaser.yaml
villevsv-upcloud: Using that instead, makes sense.

CHANGELOG.md
villevsv-upcloud: True, didn't want to mess with the tags as the v2.9.0 has been there already for a while.

kangasta (approved):

PR #242, opened 2023-07-17
feat(kubernetes): add `kubernetes nodegroup show` command

kangasta:
- Add `kubernetes nodegroup show` to display node-group details in tables. This also adds _Nodes_ table and _Anti-affinity_ field to human output.
- Show summary of cluster node-groups as table in `kubernetes show` output.
- Only accept single cluster as positional argument for node-group operations.
- Remove `(EXPERIMENTAL)` notices from kubernetes commands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add `kubernetes nodegroup show` for displaying node-group details.
+
+### Changed
+- In human readable output of `kubernetes show` command, show node-groups as table. Node-group datails are available with `kubernetes nodegroup show` command.
+
 ## [2.10.0] - 2023-07-17
 ### Added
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Add `kubernetes nodegroup show` for displaying node-group details. This also adds _Nodes_ table and _Anti-affinity_ field that were not available in previous `kubernetes show` output.
```
peknur, line 11: Not a blocker, but just wondering would `node-group` be more aligned with HTTP API resource and SDK's type names :thinking:

peknur (approved):
Looks good :+1:  Tested this locally and output looks good. Left non-blocking comment about naming convention.

CHANGELOG.md
kangasta: Yeah, we could go trough the sub-command naming. We have at least `loadbalancer` vs. `load-balancer` as well. Adding aliases would changing the naming while keeping commands backwards compatible.

PR #245, opened 2023-08-01
feat(kubernetes): add support for control-plane IP filtering

kangasta:


internal/commands/kubernetes/create.go
```diff
@@ -125,6 +125,12 @@ func (c *createCommand) InitCommand() {
 			"taint=\"env=dev:NoSchedule\","+
 			"taint=\"env=dev2:NoSchedule\"`",
 	)
+	fs.StringArrayVar(
```
kangasta, line 128: As the default value of this will be empty array the behavior from change from allow all IPs to block all IPs, which would be a breaking change.

villevsv-upcloud (approved):

PR #247, opened 2023-09-08
fix(lb): show full API response in marshaled outputs

kangasta:


internal/commands/loadbalancer/list.go
```diff
@@ -38,14 +38,17 @@ func (s *listCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Ou
...
-			{Key: "state", Header: "State", Format: format.LoadBalancerState},
+	return output.MarshaledWithHumanOutput{
+		Value: loadbalancers,
+		Output: output.Table{
+			Columns: []output.TableColumn{
+				{Key: "uuid", Header: "UUID", Colour: ui.DefaultUUUIDColours},
+				{Key: "name", Header: "Name"},
+				{Key: "plan", Header: "Plan"},
+				{Key: "zone", Header: "Zone"},
+				{Key: "operational_state", Header: "State", Format: format.LoadBalancerState},
```
kangasta, line 49: Current output uses `state` as the key, so this would be a breaking change.

villevsv-upcloud: Changelog entry to add but otherwise looks good.

villevsv-upcloud (approved):

PR #248, opened 2023-09-14
feat(dbaas): session support

villevsv-upcloud:


internal/commands/database/session/list.go
```diff
@@ -0,0 +1,187 @@
...
+
+// InitCommand implements Command.InitCommand
+func (s *listCommand) InitCommand() {
+	flagSet := &pflag.FlagSet{}
+	flagSet.IntVar(&s.limit, "limit", 10, "Number of entries to receive at most.")
+	flagSet.IntVar(&s.limit, "offset", 0, "Offset for retrieved results based on sort order.")
+	flagSet.StringVar(&s.order, "order", "query_duration:desc", "Key and direction for sorting.")
+
+	s.AddFlags(flagSet)
+	_ = s.Cobra().MarkFlagRequired("pid")
```
kangasta, line 47: Is this here by accident, cannot see a `pid` flag above?

I wonder if the ignored error would be non nil in this case 🤔 We could think of some way to check those errors at some point

internal/commands/database/session/session.go
```diff
@@ -0,0 +1,16 @@
+package databasesession
+
+import (
+	"github.com/UpCloudLtd/upcloud-cli/v2/internal/commands"
+)
+
+// BaseSessionCommand creates the base "session" command
+func BaseSessionCommand() commands.Command {
```
kangasta, line 8: If the interface is equal to removed `connection` commands, we could use that as an alias here to make this non-breaking change. Not sure if that would cause confusion, though.

villevsv-upcloud: I think it's better to keep them separate in this case. Good idea to consider in any case.

internal/commands/database/session/list.go
villevsv-upcloud: Extra flag removed.

CHANGELOG.md
```diff
@@ -12,10 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Add `--kubernetes-api-allow-ip` argument to `kubernetes create` command.
 - Add `Kubernetes API allowed IPs` field to `kubernetes show` output.
+- Add `database session list` for listing active database sessions
+- Add `database session cancel` for cancelling an active database session
 
 ### Changed
 - In human readable output of `kubernetes show` command, show node-groups as table. Node-group datails are available with `kubernetes nodegroup show` command.
 
+## Removed
+- Remove `database connection list` and `database connection cancel` commands in favor of `database session` counterparts
```
kangasta, line 22: This could have **Breaking**: in the beginning of the item, similarly as we did in v1.0.0

```suggestion
- **Breaking**: Remove `database connection list` and `database connection cancel` commands in favor of `database session` counterparts
```

PR #257, opened 2023-10-16
fix(account): pass-through API response for JSON and YAML outputs

kangasta:


internal/commands/account/show.go
```diff
@@ -80,6 +81,11 @@ func (s *showCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Ou
 			},
 		},
 	}, nil
+
+	return output.MarshaledWithHumanOutput{
```
kangasta, line 85: No change in JSON output as most recent fields (e.g., `load_balancers`) are missing from Go SDK at the moment

villevsv-upcloud (approved):

PR #266, opened 2023-11-01
feat(kubernetes): add `--version` parameter to `kubernetes create` command

kangasta:
Also adds version to output of `kubernetes show`.


peknur (approved):
Looks good :+1:

PR #268, opened 2023-11-09
feat(objectstorage): `delete`, `list` & `show` commands

villevsv-upcloud:


internal/commands/objectstorage/list.go
```diff
@@ -0,0 +1,55 @@
...
+	return output.MarshaledWithHumanOutput{
+		Value: objectstorages,
+		Output: output.Table{
+			Columns: []output.TableColumn{
+				{Key: "uuid", Header: "UUID", Colour: ui.DefaultUUUIDColours},
+				{Key: "region", Header: "Region"},
+				{Key: "configured_status", Header: "Configured status", Format: format.ObjectStorageConfiguredStatus},
+				{Key: "operational_state", Header: "Operational state", Format: format.ObjectStorageOperationalState},
+				{Key: "created_at", Header: "Created"},
+			},
```
kangasta, line 51: Updated column not defined here, but we could maybe omit also the created column to keep the table bit more concise. I guess most of the tables do not include these `created_at` and `updated_at` values either?

These only affect human output so IMO we can also modify these later once there are more important fields available.

internal/commands/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,21 @@
...
+	}
+}
+
+type objectstorageCommand struct {
+	*commands.BaseCommand
+}
+
+// InitCommand implements Command.InitCommand
+func (lb *objectstorageCommand) InitCommand() {
+	lb.Cobra().Aliases = []string{"object-storage", "objsto"}
```
kangasta, line 20: Nice 👌

internal/commands/objectstorage/objectstorage.go
```diff
@@ -0,0 +1,21 @@
...
+		commands.New("objectstorage", "Manage Managed object storage services"),
+	}
+}
+
+type objectstorageCommand struct {
+	*commands.BaseCommand
+}
+
+// InitCommand implements Command.InitCommand
+func (lb *objectstorageCommand) InitCommand() {
```
kangasta, line 19: Could use more generic or object-storage related receiver name 🙈

internal/commands/objectstorage/show.go
```diff
@@ -0,0 +1,137 @@
...
+		BaseCommand: commands.New(
+			"show",
+			"Show Managed object storage service details",
+			"upctl objectstorage show 55199a44-4751-4e27-9394-7c7661910be8",
+		),
+	}
+}
+
+type showCommand struct {
+	*commands.BaseCommand
```
kangasta, line 24: `show` and `delete` commands could have completion and resolver implementations, but those can also be added in later PR. We might also want to wait until multiple object-storages are supported per region before implementing those 🤔

internal/commands/objectstorage/delete.go
```diff
@@ -0,0 +1,42 @@
...
+		),
+	}
+}
+
+type deleteCommand struct {
+	*commands.BaseCommand
+}
+
+// Execute implements commands.MultipleArgumentCommand
+func (s *deleteCommand) Execute(exec commands.Executor, arg string) (output.Output, error) {
```
kangasta, line 27: (Unrelated) Most of our `delete` commands are quite similar, so could maybe use some generic approach for these 🤔 Created ticket about this

internal/commands/objectstorage/objectstorage.go
villevsv-upcloud: Classic. :smile: Renamed all receivers to `c` per "Command".

kangasta (approved):

PR #289, opened 2024-02-07
feat(gateway): add commands for list & delete

SCK48:


internal/commands/gateway/list.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// For JSON and YAML output, passthrough API response
+	return output.MarshaledWithHumanOutput{
+		Value: gateways,
+		Output: output.Table{
+			Columns: []output.TableColumn{
+				{Key: "uuid", Header: "UUID", Colour: ui.DefaultUUUIDColours},
+				{Key: "name", Header: "Name"},
+				{Key: "routers", Header: "Routers", Format: formatRouters},
```
kangasta, line 51: This could include features list as well, but we can add that later when we have more than `NAT` as possible value there.

internal/commands/gateway/list.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+			Rows: rows,
+		},
+	}, nil
+}
+
+func formatRouters(val interface{}) (text.Colors, string, error) {
+	routers, ok := val.([]upcloud.GatewayRouter)
+	if !ok {
+		return nil, "", fmt.Errorf("cannot parse IP addresses from %T, expected []listIPAddress", val)
```
kangasta, line 63:
```suggestion
		return nil, "", fmt.Errorf("cannot parse IP addresses from %T, expected []upcloud.GatewayRouter", val)
```

kangasta (approved):

PR #292, opened 2024-02-28
feat(objectstorage): iam changes

villevsv-upcloud:


internal/commands/objectstorage/delete.go
```diff
@@ -21,6 +24,16 @@ func DeleteCommand() commands.Command {
...
 type deleteCommand struct {
 	*commands.BaseCommand
+	deleteUsers    config.OptionalBoolean
+	deletePolicies config.OptionalBoolean
+}
+
+// InitCommand implements Command.InitCommand
+func (c *deleteCommand) InitCommand() {
+	flags := &pflag.FlagSet{}
+	config.AddToggleFlag(flags, &c.deleteUsers, "delete-users", false, "Delete users from the service.")
```
kangasta, line 34: Could maybe clarify a bit:

```suggestion
	config.AddToggleFlag(flags, &c.deleteUsers, "delete-users", false, "Delete all users from the service before deleting the object storage instance.")
```

internal/commands/objectstorage/delete.go
```diff
@@ -21,6 +24,16 @@ func DeleteCommand() commands.Command {
...
+	config.AddToggleFlag(flags, &c.deletePolicies, "delete-policies", false, "Delete policies from the service.")
```
kangasta, line 35:
```suggestion
	config.AddToggleFlag(flags, &c.deletePolicies, "delete-policies", false, "Delete all policies from the service before deleting the object storage instance.")
```

PR #297, opened 2024-04-02
feat: add paging flags for list commands that support paging

kangasta:
This also increases the default page size from 10 to 100.

### DIFF
--- a/internal/paging/paging.go
+++ b/internal/paging/paging.go
@@ -1,0 +1,23 @@
+package paging
+
+import (
+	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud/request"
+	"github.com/spf13/pflag"
+)
+
+type PageParameters struct {
+	size   int
+	number int
+}
+
+func (pp *PageParameters) ConfigureFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&pp.size, "size", 100, "Number of entries to receive at most.")
+	fs.IntVar(&pp.number, "page", 0, "Page number to calculate first item to receive. Page numbers start from `1`.")
+}
+
+func (pp *PageParameters) Page() *request.Page {
+	return &request.Page{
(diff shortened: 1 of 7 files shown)


internal/paging/paging.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+type PageParameters struct {
+	size   int
+	number int
+}
+
+func (pp *PageParameters) ConfigureFlags(fs *pflag.FlagSet) {
+	fs.IntVar(&pp.size, "size", 100, "Number of entries to receive at most.")
+	fs.IntVar(&pp.number, "page", 0, "Page number to calculate first item to receive. Page numbers start from `1`.")
```
kangasta, line 15: These could also be `--page-size` and `--page-number` to be more descriptive, but those are quite verbose 🤔

internal/paging/paging.go
```diff
@@ -0,0 +1,23 @@
+package paging
```
kangasta, line 1: Could also be `list` package if we add label filtering to the same package 🤔 Easy to rename later as well.

villevsv-upcloud: Renaming later, if needed, sounds reasonable.

internal/paging/paging.go
villevsv-upcloud: As they naturally are available only on `list` commands, these shorter ones seem fine to me.

villevsv-upcloud (approved):

internal/paging/paging.go
kangasta: I was also thinking between `--size` and `--limit`. Go SDK uses size, but there that is used as field of `Page` struct. Maybe limit would clearer without that context 🤔

kangasta: Changed `--size` → `--limit`

PR #305, opened 2024-04-24
ci(examples): ensure credentials are valid before testing examples

kangasta:


.github/workflows/examples.yml
```diff
@@ -34,4 +35,5 @@ jobs:
           UPCLOUD_PASSWORD: ${{ secrets.UPCLOUD_PASSWORD }}
         run: |
           upctl version
+          upctl account show > /dev/null # Ensure that credentials are valid
```
kangasta, line 38: If credentials are missing, this should output `Error: user credentials not found, these must be set in config file (used $HOME/.config/upctl.yaml) or via environment variables` to stderr.

PR #332, opened 2024-09-19
feat(server): take stopped/started state into account in completions

scop:
We could add some more to avoid various other completions for servers let's say in `maintenance` state, but this covers the most common cases for now.


villevsv-upcloud (approved):
Looks good to me! A short changelog entry would be useful.

I'll create a separate PR to accommodate the unrelated linter issues, so this branch can be rebased after that been merged to main.

PR #334, opened 2024-09-20
fix(linter): golangci-lint v1.61.0 compatibility

villevsv-upcloud:


internal/ui/list_layout.go
```diff
@@ -106,7 +106,7 @@ func (s *ListLayout) appendSection(title, note string, sectionBody []string) {
 		if s.style.NoteSeparator {
 			s.appendLine()
 		}
-		s.l.AppendItem(DefaultNoteColours.Sprintf(note))
+		s.l.AppendItem(DefaultNoteColours.Sprintf("%s", note))
```
kangasta, line 109: Is there `Sprint` method available? If yes, the extra _string as string_ format could be omitted 🤔

```suggestion
		s.l.AppendItem(DefaultNoteColours.Sprint(note))
```

internal/ui/ui_test.go
```diff
@@ -0,0 +1,45 @@
+package ui
+
+import (
+	"testing"
+)
+
+func TestConcatStringsSingleString(t *testing.T) {
```
kangasta, line 7: Looping through input, expected tuples and doing assert in a sub-test would also work, as all tests have the same check, but no big difference.

internal/ui/ui_test.go
```diff
@@ -0,0 +1,45 @@
...
+	result := ConcatStrings("hello")
+	expected := "hello"
+	if result != expected {
```
kangasta, line 10: `assert.Equal` helper could be used in the validations.

PR #335, opened 2024-10-04
fix(matchers): name matching case-insensitive

### LINKED ISSUE #102: Case-sensitive resource matching
**Environment (required):**
- OS: Linux
- Version 1.1.1

**Describe the bug**
It is possible to use upctl with resource names. Only problem is that the names are case-sensitive, which is an annoying UX issue. 

**To Reproduce**
```
$ upctl network list

 UUID                                   Name                    Router   Type      Zone    
────────────────────────────────────── ─────────────────────── ──────── ───────── ─────────
 03f20361-e744-4892-acf6-67d1ad1023aa   Kubernetes                       private   de-fra1 

$ upctl network show kubernetes

ERROR: cannot resolve argument: nothing found matching 'kubernetes'

$ upctl network show Kubernetes
  
  Common
    UUID:   03f20361-e744-4892-acf6-67d1ad1023aa 
    Name:   Kubernetes                           
    Router:                                      
    Type:   private                              
    Zone:   de-fra1
    etc etc
```

**Wanted behavior**
Make name-matching case-insensitive, both in name searches as well as when tab completing (can be completed to correctly cased option)

MarinX:
Fixes #102 

Added [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) to `MatchArgWithWhitespace` function name.
Added tests for `matchers.go` file.

```sh
./upctl network list                  

 UUID                                   Name              Router   Type      Zone    
────────────────────────────────────── ───────────────── ──────── ───────── ─────────
 0365d711-ab40-425e-bab3-92a1e351996d   example network            private   de-fra1 
```

```sh
./upctl network show "Example Network"
  
  Common
    UUID:   0365d711-ab40-425e-bab3-92a1e351996d 
    Name:   example network                      
    Router:                                      
    Type:   private                              
    Zone:   de-fra1                              

  Labels:

    No labels defined for this resource.
    
  IP Networks:

     Address       Family   DHCP   DHCP Def Router   DHCP DNS 
    ───────────── ──────── ────── ───────────────── ──────────
     10.0.0.0/24   IPv4     yes    no                         
    
  Servers:

     UUID                                   Title                                       Hostname             State   
    ────────────────────────────────────── ─────────────────────────────────────────── ──────────────────── ─────────
     0088f557-7939-48ec-b589-6138dbd4f084   ubuntu.example.tld (managed by terraform)   ubuntu.example.tld   started 
```

### DIFF
--- a/internal/resolver/matchers.go
+++ b/internal/resolver/matchers.go
@@ -1,10 +1,12 @@
 package resolver
 
 import (
+	"strings"
+
 	"github.com/UpCloudLtd/upcloud-cli/v3/internal/completion"
 )
 
 // MatchStringWithWhitespace checks if arg that may include whitespace matches given value. This checks both quoted args and auto-completed args handled with completion.RemoveWordBreaks.
 func MatchArgWithWhitespace(arg string, value string) bool {
-	return completion.RemoveWordBreaks(value) == arg || value == arg
+	return completion.RemoveWordBreaks(value) == arg || value == arg || strings.EqualFold(arg, value)
 }
--- a/internal/resolver/matchers_test.go
+++ b/internal/resolver/matchers_test.go
@@ -1,0 +1,43 @@
+package resolver
+
+import "testing"
+
+func TestMatchers(t *testing.T) {
+	cases := []struct {
+		name     string
+		execFn   func(string, string) bool
+		arg      string
+		value    string
+		expected bool
+	}{
+		{
+			name:     "Matcher no case",
+			execFn:   MatchArgWithWhitespace,
+			arg:      "test",
+			value:    "test",
+			expected: true,
+		},
+		{
+			name:     "Matcher with case",
+			execFn:   MatchArgWithWhitespace,
+			arg:      "TeSt",
+			value:    "test",
+			expected: true,
+		},
+		{
+			name:     "Matcher invalid",
+			execFn:   MatchArgWithWhitespace,
+			arg:      "test",
+			value:    "invalid",
+			expected: false,
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			if result := tt.execFn(tt.arg, tt.value); result != tt.expected {
+				t.Errorf("Matcher() failed %v, wanted %v", result, tt.expected)
+			}
+		})
+	}
+}


kangasta:
Thank you for the pull request! Based on quick testing in a terminal, this seems to fix the resolving part of #102. However, this would now be a breaking change for users that have resources named so that there is difference only in letter case (e.g., `McDuck` and `mcduck`), but that should be quite unlikely.

```sh
upctl router create --name mcduck
upctl router create --name McDuck
upctl router show  McDuck
# Succeeds with case-sensitive matching, fails with case-insensitive matching 
```

One option would be to use case-sensitive match in this case, but not sure how big change that would require in the resolver that calls the `MatchArgWithWhitespace` function 🤔

MarinX:
You are correct that this is breaking change. Especially for users with the same names but different casing. (unlikely but potential bug here)
What about adding a `--insensitive` flag?
```sh
upctl router show --insensitive mcduck
```
So, without a flag, the show command will function as is, while adding `--insensitive`, will trigger matching words without casing.
But, this is related only to insensitive names and its purpose is only 1 thing.

Another option is to maybe add regex to the show command.
Something like this
```sh
upctl router show --regex mc[a-z]+
```
Where we can call the regex and return the first match.
Maybe I am overengineer this too much, but thinking out loud.

Let me know your thoughts or if you have some other ideas we can discuss :)

kangasta:
Sorry about the delay with this. I'll try to continue looking into this during next week.

kangasta:
I did some testing on how the current resolver logic could be refactored to support multiple matching levels in #341. That PR adds support for using UUID prefix as argument (similarly than in e.g. `git` and `docker` commands) and the case-sensitive matching implemented in this PR could work in similar way. I'll discuss those changes with my colleague and let's merge this PR in after #341 has been merged to main.

kangasta:
#341 has now been merged, I'll rebase this on top of main early next week and will after that squash and merge this into main.

kangasta:
Thank you for the contribution @MarinX! 👍 This is now released in [v3.12.0](https://github.com/UpCloudLtd/upcloud-cli/releases/tag/v3.12.0) 🚢

MarinX:
that was fast :)
Thank you for approving my PR.

PR #348, opened 2024-12-23
feat: allow using glob pattern as an argument

kangasta:


internal/resolver/matchers.go
```diff
@@ -17,6 +18,24 @@ func MatchArgWithWhitespace(arg, value string) MatchType {
 	return MatchTypeNone
 }
 
+// MatchStringWithWhitespace checks if arg matches given value as an unix style glob pattern.
```
kangasta, line 21:
```suggestion
// MatchArgWithGlobPattern checks if arg matches given value as an unix style glob pattern.
```

Issue #83, opened 2021-05-10
[ Feature Request ] Keyring support


Davincible:
I haven't acutally used to cli yet, but reading through it the plain text credentials bare some concern to me. As a future feature request it would be great if the credentials could be stored in a keyring (be it kde, gnome, or mac etc.), and then have the cli prompt for the credentials on first use. 

Also, love the work you guys are doing here!

ajmyyra:
Hi David,

Thank you for this suggestion! It's an excellent feature and we really should add it to help people get rid of their plaintext credentials.

As for implementation, using a library with multi-OS support is kind of needed, perhaps either [99designs/keyring](https://github.com/99designs/keyring) or [zalando/go-keyring](https://github.com/zalando/go-keyring)?

kangasta:
🚢  Initial keyring support now available in [v3.16.0](https://github.com/UpCloudLtd/upcloud-cli/releases/tag/v3.16.0)

PR #350, opened 2024-12-30
feat(config): added initial keyring support (#83)

MarinX:
This PR adds initial support for the keyring.
I used [zalando/go-keyring](https://github.com/zalando/go-keyring) because of recent commits on repo (compared to [99designs/keyring](https://github.com/99designs/keyring)).
When I say initial, I mean it's open for discussion and adding more features (create auth command and save it to keyring?).

How it works:
If a username is set and the password is empty, check the keyring for service `upctl` where the username is coming either from the config file or env.
This type of logic does not break current.

Problem:
This PR does not read the username from the keyring, so it still needs a username from config or env. Maybe to introduce a config command `--keyring` ?

For me, it works by using UPCLOUD_USERNAME and creating username/password upctl services on keyring so I can switch easily between accounts without providing passwords.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -103,7 +103,9 @@ func (s *Config) Load() error {
 	if v.GetString("username") != "" && v.GetString("password") == "" {
 		password, err := keyring.Get(serviceName, v.GetString("username"))
 		if err == nil {
-			v.MergeConfigMap(map[string]interface{}{"password": password})
+			if err := v.MergeConfigMap(map[string]interface{}{"password": password}); err != nil {
+				return fmt.Errorf("unable to merge password from keyring: %w", err)
+			}
 		}
 	}
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -36,8 +36,8 @@ const (
 	// env vars custom prefix
 	envPrefix = "UPCLOUD"
 
-	// serviceName is the name of the service
-	serviceName = "upctl"
+	// keyringServiceName is the name of the service to use when using the system keyring
+	keyringServiceName = "UpCloud"
 )
 
 var (
@@ -81,7 +81,7 @@ func (s *Config) Load() error {
 	v.SetEnvPrefix(envPrefix)
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_", ".", "_"))
 	v.AutomaticEnv()
-	v.SetConfigName(serviceName)
+	v.SetConfigName("upctl")
 	v.SetConfigType("yaml")
 
 	configFile := s.GlobalFlags.ConfigFile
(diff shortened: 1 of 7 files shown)


kangasta:
Thanks for the pull request! We'll do some testing with this and let's discuss what would be the best approach to get this merged.

We are also working on adding tokens support for authentication ([product radar](https://upcloud.com/docs/product-radar/)), so probably best to start with minimal implementation, like you described above, and add more features once the initial functionality is on the main.

internal/config/config.go
```diff
@@ -95,6 +96,13 @@ func (s *Config) Load() error {
 		}
 	}
 
+	if v.GetString("username") != "" && v.GetString("password") == "" {
+		password, err := keyring.Get("upctl", v.GetString("username"))
```
scop, line 100: Just one quick nitty comment based on skimming, not a full review: let's extract the service name to a constant, and reuse in the missing_credentials.go message and tests.

internal/config/config_test.go
```diff
@@ -33,3 +34,20 @@ func TestConfig_Load(t *testing.T) {
...
 }
+
+func TestConfig_LoadKeyring(t *testing.T) {
+	cfg := New()
+	tmpFile, err := os.CreateTemp(os.TempDir(), "")
+	assert.NoError(t, err)
+	_, err = tmpFile.WriteString("username: sdkfo")
+	assert.NoError(t, err)
+
+	err = keyring.Set("upctl", "sdkfo", "foo")
```
kangasta, line 45: This will end up in the systems default keyring, so could use more descriptive username, e.g `unittest`, and a cleanup function for removing this with [t.Cleanup](https://pkg.go.dev/testing@master#T.Cleanup).

kangasta (changes requested):
Did some testing with this and seems to work nicely on mac 👍 We'll have to still confirm that this does not have conflicts with the upcoming API tokens feature.

Added one comment about the unittest and seems that at least the ubuntu test workflow needs a step for installing keyring support. I guess something like the snippet below would work for that 🤔

```yaml
- name: Install keyring test dependencies on Ubuntu
  run: apt install gnome-keyring
  if: matrix.os == 'ubuntu-latest'
```

internal/config/config_test.go
MarinX: will do. I will update the docker image and test it before pushing. Keep you posted.

MarinX: docker image updated: took a sample from [go-keyring](https://github.com/zalando/go-keyring/blob/master/.github/workflows/go.yml)
cleanup implemented: clean only the test user that was set

kangasta:
We now have the initial token support available in [v3.15.0](https://github.com/UpCloudLtd/upcloud-cli/releases/tag/v3.15.0), so I added option to read token from the keyring when no credentials are defined. I also changed the service name to `UpCloud` as we probably want to utilize the keyring in our other tooling later as well.

kangasta (approved):

villevsv-upcloud (approved):
Tested locally on Linux with Kwallet.

PR #362, opened 2025-01-29
feat: Add consistent aliases

paketeserrano:


internal/commands/util.go
```diff
@@ -141,6 +143,115 @@ func ParseSSHKeys(sshKeys []string) ([]string, error) {
...
+			return true
+		}
+	}
+	return false
+}
+
+// printDeprecationWarning prints a deprecation message
+func PrintDeprecationWarning(deprecatedAlias, newCommand string) {
+	fmt.Printf("⚠️  Deprecation Warning: The alias '%s' is deprecated and will be removed in a future release.\n", deprecatedAlias)
+	fmt.Printf("   Please use '%s' instead.\n\n", newCommand)
```
kangasta, line 239: These could be outputted to stderr to avoid problems with parsing the output.

Warning level progress messages could also be used. Those are available through `commands.Executor`. That way the formatting would be in sync with warnings we output ([here](https://github.com/UpCloudLtd/upcloud-cli/blob/beebccdb9172eb399b0bc92e391a07117fc13b02/internal/commands/executor.go#L161)) when user cancel the operation with `CTRL+C`.

paketeserrano: @kangasta I changed to print the messages to stderr. That is a good change. 
I do not understand the changes you are proposing to the progress message. The deprecation message is shown right before the command executes so I am not sure why is related to the progress of the command execution.

PR #363, opened 2025-01-30
feat(account): support API token operations

kangasta:
- [x] Depends on https://github.com/UpCloudLtd/upcloud-go-api/pull/353
- [x] Use released version of Go SDK
- [x] Update CHANGELOG

### DIFF
--- a/internal/commands/account/token/create.go
+++ b/internal/commands/account/token/create.go
@@ -76,7 +76,7 @@ func applyCreateFlags(fs *pflag.FlagSet, dst, def *createParams) {
 	fs.StringVar(&dst.expiresAt, "expires-at", def.expiresAt, "Exact time when the token expires in RFC3339 format. e.g. 2025-01-01T00:00:00Z")
 	fs.DurationVar(&dst.expiresIn, "expires-in", def.expiresIn, "Duration until the token expires. e.g. 24h")
 	fs.BoolVar(&dst.canCreateTokens, "can-create-tokens", def.canCreateTokens, "Allow token to be used to create further tokens.")
-	fs.StringArrayVar(&dst.allowedIPRanges, "allowed-ip-ranges", def.allowedIPRanges, "Allowed IP ranges for the token.")
+	fs.StringArrayVar(&dst.allowedIPRanges, "allowed-ip-ranges", def.allowedIPRanges, "Allowed IP ranges for the token. If not defined, the token can not be used from any IP. To allow access from all IPs, use `0.0.0.0/0` as the value.")
 
 	commands.Must(fs.SetAnnotation("name", commands.FlagAnnotationNoFileCompletions, nil))
 	commands.Must(fs.SetAnnotation("expires-at", commands.FlagAnnotationNoFileCompletions, nil))
--- a/internal/commands/account/token/create.go
+++ b/internal/commands/account/token/create.go
@@ -18,7 +18,7 @@ func CreateCommand() commands.Command {
 			"create",
 			"Create an API token",
 			`upctl account token create --name test --expires-in 1h`,
-			`upctl account token create --name test --expires-in 1h --allowed-ip-ranges "0.0.0.0/0" --allowed-ip-ranges "::/0"`,
+			`upctl account token create --name test --expires-in 1h --allow-ip-range "0.0.0.0/0" --allow-ip-range "::/0"`,
 		),
 	}
 }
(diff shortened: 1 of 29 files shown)


internal/commands/all/all.go
```diff
@@ -122,6 +123,13 @@ func BuildCommands(rootCmd *cobra.Command, conf *config.Config) {
 	permissionsCommand := commands.BuildCommand(permissions.BasePermissionsCommand(), accountCommand.Cobra(), conf)
 	commands.BuildCommand(permissions.ListCommand(), permissionsCommand.Cobra(), conf)
 
+	// Token
```
kangasta, line 126: These could be under account.

internal/commands/tokens/create.go
```diff
@@ -0,0 +1,120 @@
...
+func newCreateParams() createParams {
+	return createParams{
+		CreateTokenRequest: request.CreateTokenRequest{},
+	}
+}
+
+type createParams struct {
+	request.CreateTokenRequest
+	name            string
+	expiresIn       time.Duration
```
kangasta, line 43: This could use support for d unit, e.g. `90d` or `365d` could be common values defined by user 🤔

internal/commands/tokens/create.go
```diff
@@ -0,0 +1,120 @@
...
+	params  createParams
+	flagSet *pflag.FlagSet
+}
+
+func applyCreateFlags(fs *pflag.FlagSet, dst, def *createParams) {
+	fs.StringVar(&dst.name, "name", def.name, "Name for the token.")
+	fs.StringVar(&dst.expiresAt, "expires-at", def.expiresAt, "Exact time when the token expires in RFC3339 format. e.g. 2025-01-01T00:00:00Z")
+	fs.DurationVar(&dst.expiresIn, "expires-in", def.expiresIn, "Duration until the token expires. e.g. 24h")
+	fs.BoolVar(&dst.canCreateTokens, "can-create-tokens", def.canCreateTokens, "Allow token to be used to create further tokens.")
+	fs.StringArrayVar(&dst.allowedIPRanges, "allowed-ip-ranges", def.allowedIPRanges, "Allowed IP ranges for the token.")
```
kangasta, line 79: This needs more documentation as token with empty list is invalid(?) or could be made required 🤔

internal/commands/tokens/create.go
riksa: Definitely should use as hours is not convenient. I suppose one could do it with some custom parser, but the current approach uses flags.DurationVarP which uses [time.ParseDuration](https://pkg.go.dev/time#ParseDuration) and only supports up to hours.

internal/commands/all/all.go
riksa: Yeah, I was on the verge with this, so up to you. If we want to have the tokens hierarchically under accounts like they are in the API (makes sense), we should have tokens under account as a subcommand, i.e. `upctl account token create ...`.

On the other hand, it just feels better as a top level command/resource, i.e. `upctl token create ...`. 

In either case, we should follow the same hierarchy in code (style, packaging...).

kangasta: Moved the token command now under account to match API (and likely UI as well)

internal/commands/tokens/create.go
kangasta: Added this info to the help text.

internal/commands/tokens/create.go
kangasta: I think this can be improved later 🤔

paketeserrano (approved):
I am approving. Question though, should we consider releasing this as beta or experimental? Not sure if it makes sense but just in case

internal/commands/account/token/create.go
```diff
@@ -0,0 +1,125 @@
...
+	fs.StringVar(&dst.name, "name", def.name, "Name for the token.")
+	fs.StringVar(&dst.expiresAt, "expires-at", def.expiresAt, "Exact time when the token expires in RFC3339 format. e.g. 2025-01-01T00:00:00Z")
+	fs.DurationVar(&dst.expiresIn, "expires-in", def.expiresIn, "Duration until the token expires. e.g. 24h")
+	fs.BoolVar(&dst.canCreateTokens, "can-create-tokens", def.canCreateTokens, "Allow token to be used to create further tokens.")
+	fs.StringArrayVar(&dst.allowedIPRanges, "allowed-ip-ranges", def.allowedIPRanges, "Allowed IP ranges for the token. If not defined, the token can not be used from any IP. To allow access from all IPs, use `0.0.0.0/0` as the value.")
+
+	commands.Must(fs.SetAnnotation("name", commands.FlagAnnotationNoFileCompletions, nil))
+	commands.Must(fs.SetAnnotation("expires-at", commands.FlagAnnotationNoFileCompletions, nil))
+	commands.Must(fs.SetAnnotation("expires-in", commands.FlagAnnotationNoFileCompletions, nil))
+	commands.Must(fs.SetAnnotation("allowed-ip-ranges", commands.FlagAnnotationNoFileCompletions, nil))
```
villevsv-upcloud, line 84: Should we use a singular form, `allowed-ip-range`?

villevsv-upcloud (approved):
Looks good. Added a minor comment to match some other commands (plural vs singular flag)

internal/commands/account/token/create.go
kangasta: k8s command seems to use `--kubernetes-api-allow-ip` so changed this to `--allow-ip-range`.

villevsv-upcloud (approved):

PR #375, opened 2025-02-11
feat(account): add login command

kangasta:
- [ ] Sync with #350 
- [ ] Wait for hub support

### DIFF
--- a/internal/commands/account/login.go
+++ b/internal/commands/account/login.go
@@ -1,0 +1,62 @@
+package account
+
+import (
+	"fmt"
+
+	"github.com/UpCloudLtd/progress/messages"
+	"github.com/UpCloudLtd/upcloud-cli/v3/internal/commands"
+	"github.com/UpCloudLtd/upcloud-cli/v3/internal/commands/account/tokenreceiver"
+	"github.com/UpCloudLtd/upcloud-cli/v3/internal/output"
+	"github.com/zalando/go-keyring"
+)
+
+// LoginCommand creates the "account login" command
+func LoginCommand() commands.Command {
+	return &loginCommand{
+		BaseCommand: commands.New(
+			"login",
+			"Configure a authentication token.",
+			"upctl account login",
+		),
+	}
+}
+
+type loginCommand struct {
+	*commands.BaseCommand
+}
+
+// ExecuteWithoutArguments implements commands.NoArgumentCommand
+func (s *loginCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Output, error) {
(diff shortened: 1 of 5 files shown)


internal/commands/account/login.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	err = keyring.Set("UpCloud", "", token)
+	if err != nil {
+		return commands.HandleError(exec, msg, fmt.Errorf("failed to save token to keyring: %w", err))
+	}
+
+	exec.PushProgressSuccess(msg)
+
+	return output.None{}, nil
```
kangasta, line 61: Should maybe output the token if saving to keyring fails 🤔

internal/commands/account/login.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return commands.HandleError(exec, msg, err)
+	}
+
+	err = keyring.Set("UpCloud", "", token)
+	if err != nil {
+		return commands.HandleError(exec, msg, fmt.Errorf("failed to save token to keyring: %w", err))
+	}
+
+	exec.PushProgressSuccess(msg)
```
kangasta, line 59: Update progress message to something like `New authentication token created and saved to system keyring.`

internal/commands/account/login.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	token, err := receiver.Wait(exec.Context())
+	if err != nil {
+		return commands.HandleError(exec, msg, err)
+	}
+
+	err = keyring.Set("UpCloud", "", token)
+	if err != nil {
+		return commands.HandleError(exec, msg, fmt.Errorf("failed to save token to keyring: %w", err))
```
kangasta, line 56: Do we want to use config file as fallback? I guess the file should be generic to be able to use it also with e.g. Terraform provider 🤔

internal/commands/account/tokenreceiver/server.go
```diff
@@ -0,0 +1,85 @@
...
+func (s *ReceiverServer) Start() error {
+	handler := http.NewServeMux()
+	handler.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		token := req.URL.Query().Get("token")
+		if token == "" {
+			http.Redirect(w, req, getURL("error"), http.StatusSeeOther)
+			return
+		}
+		s.token = token
+		http.Redirect(w, req, getURL("success"), http.StatusSeeOther)
```
kangasta, line 45: Should the token be delivered with fetch request instead of redirects?

PR #376, opened 2025-02-13
fix: various shell flag argument completions taking fixed set of values

scop:


kangasta (approved):
Looks good to me 👌  I'd like to use backticks to highlight the values, see #382.

PR #379, opened 2025-02-21
fix: remove `v` prefix from version

kangasta:
The prefix is not used in pre-built packages. This removes it also when using go install.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -33,3 +33,31 @@ func TestConfig_Load(t *testing.T) {
 	assert.NotEmpty(t, cfg.GetString("username"))
 	assert.NotEmpty(t, cfg.GetString("password"))
 }
+
+func TestVersion(t *testing.T) {
+	tests := []struct {
+		name     string
+		expected string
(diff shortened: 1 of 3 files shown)


internal/config/config_test.go
```diff
@@ -33,3 +33,31 @@ func TestConfig_Load(t *testing.T) {
...
+			expected: "1.2.3",
+		},
+		{
+			name:     "Removes v prefix with suffix",
+			version:  "v1.2.3-rc1",
+			expected: "1.2.3-rc1",
+		},
+	}
+	for _, test := range tests {
+		test := test
```
paketeserrano, line 55: As of go 1.22 we do not need to capture the value of the variable. We do not need this assignment.

villevsv-upcloud (approved):

internal/config/config_test.go
kangasta: I guess we have more of these so we could remove these all at once 🤔

I got a linter error locally when this was not here, so I'll have to update the linter on my machine.

PR #386, opened 2025-03-04
chore(kubernetes): remove `server_number` column and hide deprecated plans

kangasta:
- [x] Depends on https://github.com/UpCloudLtd/upcloud-go-api/pull/357
- [x] Use released version of Go SDK

### DIFF
--- a/internal/commands/kubernetes/plans.go
+++ b/internal/commands/kubernetes/plans.go
@@ -37,11 +37,12 @@ func (s *plansCommand) ExecuteWithoutArguments(exec commands.Executor) (output.O
 
 	rows := []output.TableRow{}
 	for _, plan := range plans {
-		rows = append(rows, output.TableRow{
-			plan.Name,
-			plan.ServerNumber,
-			plan.MaxNodes,
-		})
+		if !plan.Deprecated {
+			rows = append(rows, output.TableRow{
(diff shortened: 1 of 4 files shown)


internal/commands/kubernetes/plans.go
```diff
@@ -37,19 +37,19 @@ func (s *plansCommand) ExecuteWithoutArguments(exec commands.Executor) (output.O
 
 	rows := []output.TableRow{}
 	for _, plan := range plans {
-		rows = append(rows, output.TableRow{
-			plan.Name,
-			plan.ServerNumber,
-			plan.MaxNodes,
-		})
+		if !plan.Deprecated {
```
paketeserrano, line 40: What if the plan is deprecated?

kangasta: The deprecated plans will be hidden from the human readable output, as now that clusters with version 1.29 cannot be deployed any more they cannot be used when deploying a cluster.

kangasta: And if there will later be deprecated plans that can be used, we could add `-a`/`--all` flag for listing those as well 🤔

villevsv-upcloud (approved):

paketeserrano (approved):

PR #393, opened 2025-03-10
feat(all): add `list` and `purge` commands

kangasta:
Initial implementation with support for networks, network-peerings, routers, databases, and object-storages.


internal/commands/all/purge.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (c *purgeCommand) InitCommand() {
+	flags := &pflag.FlagSet{}
+	flags.StringVar(&c.name, "name", "", "Only delete resources matching the given name.")
+	c.AddFlags(flags)
+}
+
+// ExecuteWithoutArguments implements commands.NoArgumentCommand
+func (c *purgeCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Output, error) {
+	msg := "Deleting all resources"
```
kangasta, line 38: Do we need confirmation prompt? I think this would mostly executed in CI cleanups with `-y` anyway, so maybe not 🤔

internal/commands/runcommand.go
```diff
@@ -139,7 +139,7 @@ func execute(command Command, executor Executor, args []string, mode resolveMode
 		argQueue = []resolvedArgument{{}}
 	}
 
-	outputs := make([]output.Output, 0, len(args))
+	outputs := make([]output.Output, 0, len(resolvedArgs))
```
kangasta, line 142: Resolved args might be longer than args, if args contained wildcards.

internal/commands/all/purge.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// PurgeCommand creates the "all purge" command
+func PurgeCommand() commands.Command {
+	return &purgeCommand{
+		BaseCommand: commands.New(
+			"purge",
+			"Delete all resources from the current account",
+			"upctl all purge",
+			"upctl all purge --include *tf-acc-test* --exclude *persistent*",
```
kangasta, line 17: This is the command we would run in Terraform acceptance test cleanup.

internal/commands/all/purge.go
villevsv-upcloud: Maybe not needed.

paketeserrano:
@kangasta Can you include tests for the new commands?

paketeserrano (approved):

PR #395, opened 2025-03-13
feat(server): add server relocation support

ucmase:


internal/commands/server/relocate.go
```diff
@@ -0,0 +1,62 @@
...
+type relocateCommand struct {
+	*commands.BaseCommand
+	completion.StoppedServer
+	resolver.CachingServer
+	zone string
+}
+
+// InitCommand implements Command.InitCommand
+func (s *relocateCommand) InitCommand() {
+	fs := &pflag.FlagSet{}
```
kangasta, line 37: Could add a long description here that briefly explains when relocating a server is possible. See [server create](https://github.com/UpCloudLtd/upcloud-cli/blob/main/internal/commands/server/create.go#L260) for example.

kangasta (approved):

PR #399, opened 2025-03-27
feat(database): add create command

### LINKED ISSUE #394: Database create command
Just to confirm, isn't `upctl database create` command available yet to be able to provision managed mysql databases from upctl?

MarinX:
This PR extends database command with create command.
Closes #394 

This is an **initial** PR that still requires some refinements:
- [x] Review flag labels to ensure consistency with other commands
- [x] Add more usage examples for better clarity and alignment with existing commands
- [x] Expand test coverage
- [x] Review property logic to determine if any adjustments are needed

Feedback is highly appreciated, especially regarding naming conventions, to ensure this PR meets standards for merging


internal/commands/database/create.go
```diff
@@ -0,0 +1,133 @@
...
+	flags := &pflag.FlagSet{}
+	s.params = createParams{CreateManagedDatabaseRequest: request.CreateManagedDatabaseRequest{}}
+	def := defaultCreateParams
+	flags.StringVar(&s.params.HostNamePrefix, "host-name-prefix", def.HostNamePrefix, "A host name prefix for the database")
+	flags.StringVar(&s.params.Title, "title", def.Title, "A short, informational description.")
+	flags.StringVar(&s.params.Plan, "plan", def.Plan, "Plan to use for the database. Run `upctl database plans [database type]` to list all available plans.")
+	flags.StringVar(&s.params.Zone, "zone", def.Zone, namedargs.ZoneDescription("database"))
+	flags.StringVar(&s.params.dbType, "type", string(def.Type), "Type of the database")
+	flags.StringSliceVar(&s.params.labels, "labels", def.labels, "Labels to describe the database in `key=value` format, multiple can be declared.\nUsage: --label env=dev\n\n--label owner=operations")
+	flags.StringSliceVar(&s.params.networks, "networks", def.networks, "A network interface for the database, multiple can be declared.\nUsage: --network family=IPv4,type=public\n\n--network type=private,network=037a530b-533e-4cef-b6ad-6af8094bb2bc,ip-address=10.0.0.1")
```
kangasta, line 92: I think we use singular for these parameters that can be defined multiple times 🤔 

```suggestion
	flags.StringSliceVar(&s.params.labels, "label", def.labels, "Labels to describe the database in `key=value` format, multiple can be declared.\nUsage: --label env=dev\n\n--label owner=operations")
	flags.StringSliceVar(&s.params.networks, "network", def.networks, "A network interface for the database, multiple can be declared.\nUsage: --network family=IPv4,type=public\n\n--network type=private,network=037a530b-533e-4cef-b6ad-6af8094bb2bc,ip-address=10.0.0.1")
```

Also, attaching network to DB is slightly different than with servers, but no need to update the example before adding parsing for the parameter. We could use similar way to parse the argument value. Another more recent example of the approach would be in `--node-group` argument of `kubernetes create` command.

internal/commands/database/create.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if len(s.properties) > 0 {
+		var props request.ManagedDatabasePropertiesRequest
+		for _, prop := range s.properties {
+			parts := strings.SplitN(prop, "=", 2)
+			if len(parts) != 2 {
+				return fmt.Errorf("invalid property format: %s, expected key=value", prop)
+			}
+			key := upcloud.ManagedDatabasePropertyKey(parts[0])
+			props.Set(key, parts[1])
```
kangasta, line 72: We probably need some parsing (like you had in the first commit) here to have correct type for the value. There are some properties that take object or array as the value, so JSON unmarshal might be simplest solution even if the command would be a bit ugly.

internal/commands/database/create.go
```diff
@@ -0,0 +1,133 @@
...
+	flags.StringVar(&s.params.HostNamePrefix, "host-name-prefix", def.HostNamePrefix, "A host name prefix for the database")
+	flags.StringVar(&s.params.Title, "title", def.Title, "A short, informational description.")
+	flags.StringVar(&s.params.Plan, "plan", def.Plan, "Plan to use for the database. Run `upctl database plans [database type]` to list all available plans.")
+	flags.StringVar(&s.params.Zone, "zone", def.Zone, namedargs.ZoneDescription("database"))
+	flags.StringVar(&s.params.dbType, "type", string(def.Type), "Type of the database")
+	flags.StringSliceVar(&s.params.labels, "labels", def.labels, "Labels to describe the database in `key=value` format, multiple can be declared.\nUsage: --label env=dev\n\n--label owner=operations")
+	flags.StringSliceVar(&s.params.networks, "networks", def.networks, "A network interface for the database, multiple can be declared.\nUsage: --network family=IPv4,type=public\n\n--network type=private,network=037a530b-533e-4cef-b6ad-6af8094bb2bc,ip-address=10.0.0.1")
+	flags.BoolVar(&s.params.terminateProtection, "terminate-protection", def.terminateProtection, "Prevents the database from being powered off, or deleted.")
```
kangasta, line 93: Could use `--termination-protection` to be consistent with GUI and API.  

```suggestion
	flags.BoolVar(&s.params.terminateProtection, "termination-protection", def.terminateProtection, "Prevents the database from being powered off, or deleted.")
```

`config.AddEnableOrDisableFlag` might also make sense here, to make this work similarly than server firewall and metadata 🤔

kangasta (commented):
Thank you for the pull request, looking good so far! I added few comments on the parameters, but have not tried to compile or further test this yet.

MarinX:
thank you for the input and suggestions. I reviewed my code and here is what I did:

- Update singular parameters name
- Use JSON marshaling for property
- Use `config.AddEnableOrDisableFlag` to be consistent as other commands
- Update on some tests to test property flag and other parameters
- Add examples to the base command

What's missing before going into full review -> network flag.

Going to check the `--node-group` flag to get a better idea of how to implement this on the database command.
Stay tuned :)

MarinX:
I think this PR is ready for a full review now!

I've added the `--maintenance` flag and completed the `--network` flag implementation. 
Instead of creating a new package, I went with a `processNetworks` function since I couldn't find an API call for creating/listing ManagedDatabaseNetworks. 
From my perspective, creating a whole package just for argument parsing felt a bit like overkill — but totally open to feedback if you see it differently! 
Maybe I missed a key point, and if needed, I can revisit the approach.

I tested several options via the terminal and everything looks solid from my end 🚀

kangasta:
On a quick glance the `processNetworks` function looks like a nice approach 👍 Sorry for the confusion with node-groups example I mentioned: `server create --network` would have been a better example after all as we do not use the same `pflag.FlagSet` in any sub-commands there (and thus do not have separate package there either). Nevertheless, looping through the key-value pairs looks like a good solution as well.

I'll continue reviewing this early next week. Have a nice weekend!

internal/commands/all/all.go
```diff
@@ -138,6 +138,7 @@ func BuildCommands(rootCmd *cobra.Command, conf *config.Config) {
 
 	// Databases
 	databaseCommand := commands.BuildCommand(database.BaseDatabaseCommand(), rootCmd, conf)
+	commands.BuildCommand(database.CreateCommand(), databaseCommand.Cobra(), conf)
```
kangasta, line 141: Because of #393 this change would need to be in `internal/commands/base/base.go` instead.

internal/commands/database/create_test.go
```diff
@@ -0,0 +1,173 @@
...
+	PrivateNetworkUUID = "03b5b0a0-ad4c-4817-9632-dafdb3ace5d9"
+	MockPrivateIPv4    = "10.0.0.1"
+	MockPrivateIPv6    = "fd42:42::1"
+	MockPublicIPv4     = "192.0.2.0"
+	MockPublicIPv6     = "2001:DB8::1"
+)
+
+func TestCreateCommand(t *testing.T) {
+	databaseDetailsMaint := upcloud.ManagedDatabase{
+		UUID: UUID1,
```
kangasta, line 30: Seems that only `UUID1` is used, so other vars could be removed from here. Maybe even inline `UUID1` since it is only used once 🤔 

```suggestion
func TestCreateCommand(t *testing.T) {
	databaseDetailsMaint := upcloud.ManagedDatabase{
		UUID: "0927dfd6-3884-4079-a948-3a8881df1a7a",
```

kangasta (approved):
On overall looks good to me! Added note about unused vars in test file and naming of maintenance related parameters. This will need merge conflict fix as well, but that would be just moving `BuildCommand` call from `all` package to `base` package.

internal/commands/database/create.go
```diff
@@ -0,0 +1,214 @@
...
+	flags := &pflag.FlagSet{}
+	s.params = createParams{CreateManagedDatabaseRequest: request.CreateManagedDatabaseRequest{}}
+	def := defaultCreateParams
+	flags.StringVar(&s.params.HostNamePrefix, "host-name-prefix", def.HostNamePrefix, "A host name prefix for the database")
+	flags.StringVar(&s.params.Title, "title", def.Title, "A short, informational description.")
+	flags.StringVar(&s.params.Plan, "plan", def.Plan, "Plan to use for the database. Run `upctl database plans [database type]` to list all available plans.")
+	flags.StringVar(&s.params.Zone, "zone", def.Zone, namedargs.ZoneDescription("database"))
+	flags.StringVar(&s.params.dbType, "type", string(def.Type), "Type of the database")
+	flags.StringVar(&s.params.Maintenance.DayOfWeek, "dow", def.Maintenance.DayOfWeek, "Full name of weekday in English, lower case(sunday) for automatic maintenance day of the week. Set randomly if not provided.")
+	flags.StringVar(&s.params.Maintenance.Time, "time", def.Maintenance.Time, "Database time in UTC of automatic maintenance HH:MM:SS. Set randomly if not provided.")
```
kangasta, line 170: and maybe these two could have `maintenance-` prefix to be more self explaining 🤔 

```suggestion
	flags.StringVar(&s.params.Maintenance.DayOfWeek, "maintenance-dow", def.Maintenance.DayOfWeek, "Full name of weekday in English, lower case(sunday) for automatic maintenance day of the week. Set randomly if not provided.")
	flags.StringVar(&s.params.Maintenance.Time, "maintenance-time", def.Maintenance.Time, "Database time in UTC of automatic maintenance HH:MM:SS. Set randomly if not provided.")
```

paketeserrano:
@MarinX , there is an issue parsing the version number. You can see the problem here with a few traces:
`paco@nube:~/github/upcloud-cli/bin$ ./upctl database create --title=my-db --zone=es-mad1 --type=pg --hostname-prefix=lumni-db-12345 --enable-termination-protection --plan=2x4xCPU-8GB-50GB --network name=production-primary-db,family=IPv4,type=private,uuid=03b52674-10e3-4254-9d77-2479bc3d9c9d --property public_access=false --property version="17" --property password_encryption="scram-sha-256" --wait
Before unmarshal: key=public_access, value=false
Parsed value: key=public_access, value=false
Before unmarshal: key=version, value=17
Parsed value: key=version, value=17
Before unmarshal: key=password_encryption, value=scram-sha-256
Unmarshal failed, using plain string: scram-sha-256
✗ Creating database my-db                                                                                                                                                               
  Error: error: message="Validation error.", type="https://developers.upcloud.com/1.3/errors#ERROR_INVALID_REQUEST", correlation_id=01JWB12NG7YPX35HMBSZQD4S3V,                         
  invalid_params_properties.version='Expected string or null, but got number.'

Error: Command execution failed for 1 resource(s)
`
It is related to this code:
`func processProperties(in []string) (request.ManagedDatabasePropertiesRequest, error) {
	resp := request.ManagedDatabasePropertiesRequest{}
	for _, prop := range in {
		parts := strings.SplitN(prop, "=", 2)
		if len(parts) != 2 {
			return resp, fmt.Errorf("invalid property format: %s, expected key=value", prop)
		}

		key := upcloud.ManagedDatabasePropertyKey(parts[0])
		value := parts[1]

		// remove quotes from the start and end of the value if they exist
		if (strings.HasPrefix(value, "\"") && strings.HasSuffix(value, "\"")) ||
			(strings.HasPrefix(value, "'") && strings.HasSuffix(value, "'")) {
			value = value[1 : len(value)-1]
		}

		fmt.Printf("Before unmarshal: key=%s, value=%s\n", key, value)

		var parsedValue interface{}
		if err := json.Unmarshal([]byte(value), &parsedValue); err != nil {
			fmt.Printf("Unmarshal failed, using plain string: %s\n", value)
			resp.Set(key, value) // set as plain string if parsing fails
		} else {
			fmt.Printf("Parsed value: key=%s, value=%#v\n", key, parsedValue)
			resp.Set(key, parsedValue)
		}
	}
	return resp, nil
}`

The api is expecting a string in the version field. Do you see a nice way of fixing this one?

paketeserrano:
Hey @MarinX ,

I actually made some testing and I did not see the need for stripping the quotes. You can see some results here:

https://app.asana.com/1/6339747423975/project/1206688814028894/task/1210383513264708?focus=true

I am leaning towards giving the user to set proper json format and documenting it properly. I created this PR:
https://github.com/UpCloudLtd/upcloud-cli/pull/422

Let me know if this sounds reasonable for you or if you see any other caveat.

MarinX:
hey @paketeserrano , makes sense.
After testing looks like I added the removal of quotes related to shell input.
For example, `key="true"` will be string "true" while `key=true` will be boolean true. 
By adding the removal of quotes, it will always return true.

This "smart" logic bites back when you want to have a string "17" instead of number and causes the issue.

Glad you figured out the solution, and it looks valid to me. :+1:

paketeserrano:
Thanks for checking it out, appreciate it @MarinX

PR #410, opened 2025-04-17
fix(runcommand): output correct invalid creds error

### LINKED ISSUE #392: Suboptimal error message on token delete with invalid credentials
**Environment (required):**
- OS: [e.g. macOS (Darwin or M1), Linux, Windows] Linux
- Version [e.g. 1.2.0, "upctl version" gives you this] 3.16.1

**Describe the bug**
`upctl account token delete THING` error message is not as pretty as it c/should when accessed using invalid credentials.

**To Reproduce**

1. Configure invalid credentials
2. `upctl account token delete FOO`

**Expected behavior**
```shellsession
$ upctl account token delete FOO
✗ Command execution failed                                                                                                                                                     
  Error: error: message="Authentication failed using the given username and password.", type="AUTHENTICATION_FAILED"
```

**Observed behavior**
```shellsession
$ upctl account token delete FOO
Error: cannot resolve command line arguments: cannot get resolver: error: message="Unauthorized.", type="https://developers.upcloud.com/1.3/errors#ERROR_UNAUTHORIZED", correlation_id=...
```

**Screenshots**
N/A

**Additional context**
N/A

MarinX:
closes #392 

The issue:
Invalid creds returns error at [runcommand.go](https://github.com/UpCloudLtd/upcloud-cli/blob/e7fda070e023cee2c674760d1d058bbb1b0b5575/internal/commands/runcommand.go#L119) only if resolver is used.
If the resolver is not used (ex., server list), it will go into the worker queue and render an error with error output. 
This is why some commands return expected output and some just an error message.

The fix:
output the error and stop progress log to print error message as other commands do without the resolver.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Set exit code to 103 if authentication fails also in commands that do not take arguments (e.g. `server list`).
+
 ## [3.19.1] - 2025-04-29
 
 ### Fixed
--- a/internal/commands/runcommand.go
+++ b/internal/commands/runcommand.go
@@ -80,7 +80,14 @@ const (
 
 func resolveArguments(nc Command, exec Executor, args []string, mode resolveMode) (out []resolvedArgument, err error) {
 	if mode == resolveNone {
-		return nil, nil
+		acc := exec.Account()
+		if acc == nil {
+			// we don't have account service, nothing to validate
+			return nil, nil
+		}
+		// we only care about error here to report back any issue with auth
+		_, err := acc.GetAccount(exec.Context())
+		return nil, err
 	}
 
 	if resolve, ok := nc.(resolver.ResolutionProvider); ok {
--- a/internal/commands/runcommand_test.go
+++ b/internal/commands/runcommand_test.go
@@ -188,6 +188,7 @@ func TestRunCommand(t *testing.T) {
 				test.command.(*mockMulti).On("Execute", mock.Anything, mock.Anything).Return(output.OnlyMarshaled{Value: "mock"}, nil)
 			}
 			mService := &smock.Service{}
+			mService.On("GetAccount").Return(nil, nil)
 			cfg := config.New()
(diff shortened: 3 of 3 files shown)


kangasta:
Created #411 for this as I think #392 is more about the content of the error message than its formatting.

MarinX:
Appreciate the response and suggested fix.
Looks like it's a better approach than just erroring out and returning the wrong status code.

Let me work on the PR and get back with the changes.

In the meantime, if someone else wants to work on it, just let me know so we don't do double work.
A little bit slow this week as other things got in the way (sorry) :)

MarinX:
Seems like there are multiple paths to fixing this issue.

### Option 1.
By adding a resolver to every service (I added just for an account), the resolver can be called and exit with the correct error code.
Since there are a lot of services, this PR can potentially grow big with changes, as every service needs a resolver. 
For the services without a resolver ([sample in account/permission/list](https://github.com/UpCloudLtd/upcloud-cli/pull/410/files#diff-80779fb819ee86f354b808c4c593b8d3be83893879802d6db70228f8614373ebR29)), should we default to CachingAccount resolver?

### Option 2
Most straightaway solution. 
[If the resolve mode is none, execute a GetAccount() and return err](https://github.com/UpCloudLtd/upcloud-cli/pull/410/files#diff-fd87c5a010d55f414aa782f4cbbb0f9708a948a2dd0b67f58b203e81597ab1d7R83).
While it is a "single line" fix, tests need to be updated and can cause a debugging headache because of the extra "hidden" call for a GetAccount.

Can I get your input on this @kangasta ?

MarinX:
updated PR to call GetAccount if the resolve mode is none

```sh
➜  upctl git:(fix/execute-error) ✗ ./upctl account show
Error: invalid user credentials, authentication failed using the given username and password
➜  upctl git:(fix/execute-error) ✗ echo $?
103
➜  upctl git:(fix/execute-error) ✗ ./upctl server list
Error: invalid user credentials, authentication failed using the given username and password
➜  upctl git:(fix/execute-error) ✗ echo $?
103
➜  upctl git:(fix/execute-error) ✗ ./upctl version
  
  Version:      dev-9d0ec785-dirty 
  Build date:   unknown            
  Built with:   go1.24.3           
  System:       linux              
  Architecture: amd64   
  ```

PR #420, opened 2025-05-23
docs(README): remove `go install` instructions

scop:
This may not be the recommended way to install, and the previous getting started chapter already links to the complete install instructions, including from prebuilt binaries (as well as the `go install` way).


thevilledev (approved):
LGTM

villevsv-upcloud (approved):

PR #422, opened 2025-05-28
Feat(property parsing): version property

paketeserrano:


MarinX:
test fail on [create_test.go](https://github.com/UpCloudLtd/upcloud-cli/blob/338974fdafc22e3281cba0985694799e9698b5e7/internal/commands/database/create_test.go#L71)

Since the logic of the quote is removed, this is treated as a string instead of an object
`\"{\"client_id\":\"test_client_id\"}\"`
removing the extra quotes will make the test pass, like this:
`{\"client_id\":\"test_client_id\"}`

kangasta:
I forgot these strings that look like numbers in earlier review 😅 There might be some other problematic types/values as well.

I guess we need some logic to define the type based on [database type details](https://developers.upcloud.com/1.3/16-managed-database/#get-managed-database-type-details) response.

PR #431, opened 2025-06-23
fix: do not run as root in container

scop:


kangasta:
Could add CHANGELOG entry about this, otherwise looks good to me! 👍

PR #465, opened 2025-07-08
feat: audit log export support 

scop:
Requires https://github.com/UpCloudLtd/upcloud-go-api/pull/377 and updated to a new upcloud-go-api release containing it.

### DIFF
--- a/internal/commands/auditlog/export.go
+++ b/internal/commands/auditlog/export.go
@@ -1,0 +1,44 @@
+package auditlog
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/UpCloudLtd/upcloud-cli/v3/internal/commands"
+	"github.com/UpCloudLtd/upcloud-cli/v3/internal/output"
+	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud/request"
+	"github.com/spf13/pflag"
+)
+
+var formats = []string{
+	request.ExportAuditLogFormatJSON,
+	request.ExportAuditLogFormatCSV,
+}
+
+// ExportCommand creates the "audit-log export" command
+func ExportCommand() commands.Command {
+	return &exportCommand{
+		BaseCommand: commands.New("export", "Export audit logs", "upctl audit-log export", "upctl audit-log export --output csv >audit-log.csv"),
+	}
+}
+
+type exportCommand struct {
+	*commands.BaseCommand
+	params request.ExportAuditLogRequest
+}
+
+func (c *exportCommand) InitCommand() {
+	fs := &pflag.FlagSet{}
+	fs.StringVar(&c.params.Format, "output", formats[0], fmt.Sprintf("Export format (typically %s). Note that this overrides the global --output flag.", strings.Join(formats, "|")))
+	c.AddFlags(fs)
+}
+
+// ExecuteWithoutArguments implements [commands.NoArgumentCommand].
+func (c *exportCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Output, error) {
+	r, err := exec.All().ExportAuditLog(exec.Context(), &c.params)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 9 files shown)


internal/output/render.go
```diff
@@ -45,10 +49,15 @@ func Render(writer io.Writer, outputFormat string, commandOutputs ...Output) (er
...
-	// Count failed outputs
+	// Render streaming outputs and count failed ones
 	failedCount := 0
 	for _, commandOutput := range commandOutputs {
-		if _, ok := commandOutput.(Error); ok {
+		if rawOutput, ok := commandOutput.(Raw); ok {
+			defer rawOutput.Close()
+			if _, err := io.Copy(writer, rawOutput); err != nil {
+				return err
+			}
```
scop, line 59: I'm not too happy with this implementation, but at least it is a rather non-intrusive one.

internal/commands/auditlog/export.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type exportCommand struct {
+	*commands.BaseCommand
+	params request.ExportAuditLogRequest
+}
+
+func (c *exportCommand) InitCommand() {
+	fs := &pflag.FlagSet{}
+	fs.StringVar(&c.params.Format, "output", formats[0], fmt.Sprintf("Export format (typically %s). Note that this overrides the global --output flag.", strings.Join(formats, "|")))
```
scop, line 32: Overriding the global `--output` this way is a bit of a hack, but avoids adding another formatting option or adding awkward ones that won't work everywhere to the global.

internal/commands/base/base.go
```diff
@@ -230,6 +231,10 @@ func BuildCommands(rootCmd *cobra.Command, conf *config.Config) {
 	commands.BuildCommand(partneraccount.CreateCommand(), partnerAccountCommand.Cobra(), conf)
 	commands.BuildCommand(partneraccount.ListCommand(), partnerAccountCommand.Cobra(), conf)
 
+	// Audit log operations
+	auditlogCommand := commands.BuildCommand(auditlog.BaseAuditLogCommand(), rootCmd, conf)
```
kangasta, line 235: Could also have this under account command similarly than we have audit-logs page under account details in hub, but no strong opinions on this from my side 🤔

internal/commands/auditlog/export.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// ExecuteWithoutArguments implements [commands.NoArgumentCommand].
+func (c *exportCommand) ExecuteWithoutArguments(exec commands.Executor) (output.Output, error) {
+	r, err := exec.All().ExportAuditLog(exec.Context(), &c.params)
+	if err != nil {
+		return nil, err
+	}
+
+	return output.Raw{Source: r}, nil
```
kangasta, line 43: Seems that global interrupt signal handling is not working when this command is streaming content to console: `panic: send on closed channel`

internal/commands/auditlog/export.go
```diff
@@ -0,0 +1,44 @@
...
+
+var formats = []string{
+	request.ExportAuditLogFormatJSON,
+	request.ExportAuditLogFormatCSV,
+}
+
+// ExportCommand creates the "audit-log export" command
+func ExportCommand() commands.Command {
+	return &exportCommand{
+		BaseCommand: commands.New("export", "Export audit logs", "upctl audit-log export"),
```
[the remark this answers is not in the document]
scop: Added.

internal/commands/auditlog/export.go
scop: Good catch. This is not specific to this change, but in other commands besides this one the interrupt window is much smaller and thus less likely gets hit. Anyway, https://github.com/UpCloudLtd/upcloud-cli/pull/503 should address it.

villevsv-upcloud: Rebased, #503 seems to have fixed the issue :+1:

PR #547, opened 2025-10-02
feat(stack): add destroy command

paketeserrano:


internal/commands/stack/stackops.go
```diff
@@ -316,35 +316,56 @@ func DestroyStack(exec commands.Executor, name, zone string, deleteStorage, dele
...
+			if storageUUID == "" {
+				exec.PushProgressUpdateMessage(msg, fmt.Sprintf("Object storage %s not found, skipping deletion", objectStorageName))
+				return nil
+			}
+
+			if deleteObjectStorage {
+				_, err = objectstorage.Delete(exec, storageUUID, true, true, true, true)
+				if err != nil {
+					return fmt.Errorf("failed to delete object storage: %w", err)
+				}
```
villevsv-upcloud: We could return nil after the error check, then the else would not be needed.

PR #560, opened 2025-10-20
chore(snyk): remove superfluous MPL-2.0 ignores

scop:


villevsv-upcloud:
> I thought the snyk job was failing because of this mlp 2.0 license. Is this not a problem anymore for snyk?

MPL-2.0 license is allowed now, so we can remove this ignore. :+1:

PR #607, opened 2025-11-28
feat(debug): log API requests and responses when using `--debug`

kangasta:


internal/config/config.go
```diff
@@ -206,6 +206,8 @@ func (s *Config) CreateService() (internal.AllServices, error) {
 	password := s.GetString("password")
 	token := s.GetString("token")
 
+	apiLogger := flume.New("api")
```
kangasta: Didn't yet test how this works with progress logging. I guess we could do progress logging in file mode if `--debug` is enabled, if debug logging causes problems with updating active progress lines 🤔

paketeserrano (approved):

PR #637, opened 2025-12-16
docs: add missing policy attachment step for object storage

### LINKED ISSUE #570: Object storage documentation is missing attach policy step
**Is your feature request related to a problem? Please describe.**
Following these guides:   
https://upcloudltd.github.io/upcloud-cli/examples/use_object_storage/   https://upcloud.com/docs/guides/connecting-to-an-object-storage-instance/. 

I couldn't get access to my newly created bucket using the AWS CLI.  
I could aws s3 ls after attaching an ECSS3FullAdmin policy in the console.  

**Describe the solution you'd like**
A complete E2E guide on how to get access to a bucket usign upctl and aws CLI.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

mgajda:
Fixes #570

Adds the missing step for attaching user access policies to object storage users. Without this step, users cannot access buckets via AWS CLI even with valid S3 credentials.

## Changes
- Add policy attachment via UpCloud API (curl with bearer token)
- Add policy attachment via web console (alternative)
- Add S3 access verification example
- Clarify difference between UpCloud API credentials and S3 access keys

## Testing
✅ End-to-end workflow tested
✅ API endpoint verified (HTTP 204 success)
✅ Credential types clarified

This directly addresses the issue reported by @erikologic where the policy attachment step was missing from the documentation.

### DIFF
--- a/examples/use_object_storage.md
+++ b/examples/use_object_storage.md
@@ -51,7 +51,7 @@ Attach a policy to grant the user access to buckets:
    ```
 
    Then attach the policy (requires UPCLOUD_TOKEN environment variable):
-   ```sh when=false
+   ```sh when="false"
    # Note: This command requires a bearer token which can be created via the UpCloud Control Panel
    # The when=false flag skips this in automated tests since only username/password are available in CI
    curl -X POST "https://api.upcloud.com/1.3/object-storage-2/${service_uuid}/users/${prefix}user/policies" \
@@ -69,7 +69,7 @@ Attach a policy to grant the user access to buckets:
 
 Verify S3 access with AWS CLI:
 
-```sh when=false
+```sh when="false"
 # Get the service endpoint
 service_endpoint=$(upctl object-storage show ${prefix}service -o json | jq -r '.endpoints[0].domain_name')
 
@@ -79,13 +79,7 @@ AWS_SECRET_ACCESS_KEY=${secret_access_key} \
 aws s3 ls --endpoint-url https://${service_endpoint}
 ```
 
-Once not needed anymore, delete the user:
-
-```sh
-upctl object-storage user delete ${prefix}service --username ${prefix}user
-```
-
-Delete also the managed object storage service along with all its sub-resources such as buckets and users:
+Delete the managed object storage service along with all its sub-resources such as buckets and users:
 
 ```sh
 upctl object-storage delete ${prefix}service --force


examples/use_object_storage.md
```diff
@@ -35,6 +35,44 @@ upctl object-storage user create ${prefix}service --username ${prefix}user
...
+   curl -X POST "https://api.upcloud.com/1.3/object-storage-2/<service-uuid>/users/${prefix}user/policies" \
+     -H "Authorization: Bearer ${UPCLOUD_TOKEN}" \
+     -H "Content-Type: application/json" \
+     -d '{"name": "ECSS3FullAccess"}'
+   ```
+
+   A successful response returns HTTP status 204.
+
+2. **Using the UpCloud Control Panel:**
+   Navigate to Object Storage → Users → Select user → Attach Policy → ECSS3FullAccess
```
kangasta: Instead of these, we would want to have example on how to do this `upctl`. Problem is that we do not currently have commands to do that: #647

examples/use_object_storage.md
```diff
@@ -35,6 +35,44 @@ upctl object-storage user create ${prefix}service --username ${prefix}user
...
+
+**Note:** Without attaching a policy, the user won't have permission to access buckets via AWS CLI or S3-compatible tools.
+
+Verify S3 access with AWS CLI:
+
+```sh
```
kangasta: We use these code snippets as scripts when testing `upctl` (see README in examples directory for more info), so I think this would fail because we do not have `aws` command available. Could either check if we can use curl to test the connection in somewhat nice way, or skip the code-block with `when=false`:

```suggestion
```sh when=false
```

examples/use_object_storage.md
mgajda: @kangasta I agree having object storage managed by `upctl` is desirable. But I would leave current example as-is, until the new commands are implemented. This ensures users and tests reflect the current state, instead of future plans.

After we merge this, I would happily propose another PR for #647 and additional object storage commands to allow flexible management of object storage.

paketeserrano:
@mgajda this is a good addition since it makes the obj storage example complete covering how to enable a user to access a bucket.

These type of files are also used as tests in the workflow .github/workflows/examples.yaml and it will fail as it is. Would it be possible to add the commands to parse the service-uuid, your-access-key-id, your-secret-access-key and service-endpoint ? That would make the test that executes the shell commands in this file succeeds.

Issue #643, opened 2025-12-19
CLI login with token fails on Windows 11 ARM WSL


eliasylanen:
**Environment (required):**
- Computer: Microsoft Surface 7 ARM
- OS: Ubuntu 24.04 on Windows 11 WSL
- Version 3.26.0

**Describe the bug**
```sh
elias@LAPTOP-7VLOTAIM:~ $ echo "$TOKEN" | upctl account login --with-token
✗ Saving provided token to the system keyring.
  Error: failed to unlock correct collection '/org/freedesktop/secrets/aliases/default'
Error: Command execution failed for 1 resource(s)
```

**Expected behavior**
The login succeeds and `~/.config/upctl.yml` is created

**Observed behavior**
The aforementioned error, and the config file remains uncreated

**Additional context**
Running WSL on a Windows ARM laptop, so the reason could lie with either one 😅
The exact same token and command has been confirmed working on plain AMD64 Fedora.

mgajda:
@eliasylanen Did you try to unlock keyring first using this guide: https://github.com/jaraco/keyring#using-keyring-on-headless-linux-systems

It seems that WSL workarounds may harm security, and WSL keyring may not work for awhile: https://askubuntu.com/questions/815327/running-gnome-keyring-on-wsl-windows-subsystem-for-linux.

I suggest adding a `--no-keyring` option that prevents crash.

kangasta:
There should be no need for `--no-keyring` flag: keyring has least priority when parsing the credentials, so with `--no-keyring` user would just get a slightly different error message. I.e., if credentials are defined in environment variables, the keyring should not be accessed. Maybe instead of adding `--no-keyring` flag the keyring access failed error message could be clarified.

For `login --with-token`, the `--no-keyring` flag would not make sense as we do not currently have any alternative methods for saving the token. One alternative would be to save the token to upctl configuration file, but in that case the parameter should be about allowing this less secure fallback.

eliasylanen:
Hi! I've been busy with work for the last few months, so haven't been able to return to this. I've also switched to full-on Linux, where this all just works, so I sadly won't be able to debug or test this any further on my end

PR #653, opened 2025-12-31
feat(auth): add --no-keyring flag to address keyring access issues (#643)

mgajda:
## Summary
Implements `--no-keyring` flag to address issue #643 where users on WSL and ARM systems experience keyring access failures.

## Changes
- Added `--no-keyring` global flag to disable system keyring integration
- Added support for `no-keyring` configuration option in `upctl.yaml ` 
- Modified `account login` command to save tokens to config file when keyring is disabled
- Skipped `credentials.Parse()` calls when keyring is disabled to avoid access attempts
- Added helpful error messages suggesting `--no-keyring` when keyring operations fail

## Motivation
Users on WSL, ARM, and other environments often encounter keyring errors like:
```
Error: failed to unlock correct collection '/org/freedesktop/secrets/aliases/default'
```

This PR allows them to bypass the keyring entirely and use configuration file storage instead.

## Usage
```bash
# Use --no-keyring flag
upctl account login --with-token --no-keyring
```
Or set in config file:
```bash
echo "no-keyring: true" >> ~/.config/upctl.yaml
```
Works with all commands
```bash
upctl server list --no-keyring
```

## Testing
- Tested login with `--no-keyring` flag saves token to config file
- Tested config file option `no-keyring: true` works
- Verified normal keyring flow still works when flag is not set
- Confirmed helpful error hints are shown on keyring failures

For #643

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -61,7 +61,7 @@ type GlobalFlags struct {
 	OutputFormat  string        `valid:"in(human|json|yaml)"`
 	NoColours     OptionalBoolean
 	ForceColours  OptionalBoolean
-	NoKeyring     bool          `valid:"-"`
+	NoKeyring     bool `valid:"-"`
 }
 
 // Config holds the configuration for running upctl
@@ -275,10 +275,10 @@ func WrapKeyringError(err error) error {
 	// Check for common keyring error patterns
 	errStr := err.Error()
 	if strings.Contains(errStr, "failed to unlock") ||
-	   strings.Contains(errStr, "keyring") ||
-	   strings.Contains(errStr, "secret") ||
-	   strings.Contains(errStr, "dbus") ||
-	   strings.Contains(errStr, "collection") {
+		strings.Contains(errStr, "keyring") ||
+		strings.Contains(errStr, "secret") ||
+		strings.Contains(errStr, "dbus") ||
+		strings.Contains(errStr, "collection") {
 		return fmt.Errorf("%w\n\nHint: If you're experiencing keyring issues, you can:\n  1. Use the --no-keyring flag to save credentials to config file instead\n  2. Set 'no-keyring: true' in your upctl.yaml config file\n  3. Use environment variables UPCLOUD_TOKEN for authentication", err)
 	}
 


internal/core/core.go
```diff
@@ -102,6 +102,10 @@ func BuildRootCmd(conf *config.Config) cobra.Command {
 		&conf.GlobalFlags.Debug, "debug", false,
 		"Print out more verbose debug logs.",
 	)
+	flags.BoolVar(
+		&conf.GlobalFlags.NoKeyring, "no-keyring", false,
+		"Disable system keyring integration for credential storage.",
+	)
```
kangasta: _Comment from #643 to make it visible for other reviewers:_

> There should be no need for `--no-keyring` flag: keyring has least priority when parsing the credentials, so with `--no-keyring` user would just get a slightly different error message. I.e., if credentials are defined in environment variables, the keyring should not be accessed. Maybe instead of adding `--no-keyring` flag the keyring access failed error message could be clarified.
>
> For `login --with-token`, the `--no-keyring` flag would not make sense as we do not currently have any alternative methods for saving the token. One alternative would be to save the token to upctl configuration file, but in that case the parameter should be about allowing this less secure fallback.

kangasta (changes requested):

internal/commands/account/login.go
```diff
@@ -57,14 +58,42 @@ func (s *loginCommand) executeWithToken(exec commands.Executor) (output.Output,
...
+			v.SetConfigFile(configFile)
+			v.SetConfigType("yaml")
+			_ = v.ReadInConfig()
+			noKeyring = v.GetBool("no-keyring")
+		}
+	}
+	if noKeyring {
+		// Get config file path from flag if specified
+		configFile, _ := s.Cobra().Root().PersistentFlags().GetString("config")
+		msg := "Saving token to configuration file (keyring disabled)."
```
kangasta: We use the token saved with upctl in our other tooling as well (see [credentials](https://github.com/UpCloudLtd/upcloud-go-api/tree/main/credentials) helper), so this should be saved in more generic location (e.g. UpCloud level config instead of UpCloud CLI level) 🤔

internal/commands/account/login.go
```diff
@@ -57,14 +58,42 @@ func (s *loginCommand) executeWithToken(exec commands.Executor) (output.Output,
...
+		v := viper.New()
+		configFile, _ := s.Cobra().Root().PersistentFlags().GetString("config")
+		if configFile != "" {
+			v.SetConfigFile(configFile)
+			v.SetConfigType("yaml")
+			_ = v.ReadInConfig()
+			noKeyring = v.GetBool("no-keyring")
+		}
+	}
+	if noKeyring {
```
kangasta: Instead of checking for `--no-keyring` we could write to a config file if keyring access failed because of a specific error. In that case an warning should be printed.

mgajda:
Replaced by #654 after feedback.

PR #656, opened 2026-01-02
chore(deps): configure Dependabot cooldown

scop:
Refs
- https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-
- https://docs.zizmor.sh/audits/#dependabot-cooldown


.github/dependabot.yml
scop: As a general note, we could consider excluding at least some deps of our own from cooldown, https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#configuration-of-cooldown

The reasons for including them in cooldown is the same as for other dependencies, but for example changes in upcloud-cli do require changes made in upcloud-go-api to be brought in at times, and Dependabot could help out a bit more aggressively wrt those. Then again it's not that much of a biggie to take care of the dependency updates ourselves for cases like that. (Leaned towards that myself and therefore did not add any excludes.)

